PR #29, opened 2022-10-24
Make WASM download a no-brainer

sourishkrout:
If GitHub's `gh` CLI is installed. If not, it'll just prompt for the token.


christian-bromann (approved):
👍

PR #34, opened 2022-10-25
Vercel Integration

christian-bromann:
This patch adds a similar deployment workflow as we have for Deno. It reads the Vercel command and runs the job. As oppose to a pure CLI execution, this one uses their API to better integrate into the VS Code environment.

Demo:
RunmeVercelOptim


sourishkrout (changes requested):
Pending out of band discussion.

sourishkrout (approved):
Merging.

PR #35, opened 2022-10-25
Render non supported languages as markdown

christian-bromann:
This patch is a proposal to just have non supported languages being rendered as plain markdown, e.g.:

Screenshot 2022-10-24 at 19 45 47


sourishkrout (approved):
Yep, LGTM 👍

PR #38, opened 2022-10-26
Allow user to determine working directory

### LINKED ISSUE #25: CLI button won't pass subdirectory/filename
If a markdown file is located in a subdirectory, the CLI button will launch `runme run` in the incorrect working directory.

christian-bromann:
fixes #25

The current implementation always runs tasks/shells based on the directory where the opened markdown is located. Users might however like to always work based on the working directory of the workspace. This patch allows to define this with having "Relative to Markdown" as default.

### DIFF
--- a/package.json
+++ b/package.json
@@ -128,6 +128,16 @@
             "scope": "machine",
             "default": true,
             "markdownDescription": "If set to `true`, the terminal closes if task succeeds running."
+          },
+          "runme.shell.workingDirectory": {
+            "type": "string",
+            "scope": "machine",
+            "default": "Relative to Markdown File",
+            "enum": [
+              "Relative to Markdown File",
+              "Relative to Workspace Directory"
+            ],
+            "markdownDescription": "Used working directory for shell or task executions."
           }
         }
       }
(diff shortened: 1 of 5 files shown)


sourishkrout:
> fixes #25
> 
> The current implementation always runs tasks/shells based on the directory where the opened markdown is located. Users might however like to always work based on the working directory of the workspace. This patch allows to define this with having "Relative to Markdown" as default.

The issue described in #25 is ever so slightly different where I don't see a user setting to be the correct solution. Is there any other bug/issue that prompted this implementation?

Outside of bug #25, I don't think we should put it on the user to adjust to the correct working directory. If there's a sufficient reason we could add a `{ cwd=./relative/path }` to the attributes will apply to a single cell and goes along with providing control to the author/owner/maintainer.

sourishkrout:
I'll play around with this. While it's not a bugfix, it might helpful for other scenarios. Leaving it open for now.

sourishkrout:
Hasn't come up as a hands-on problem. Can revive if we'll receive user reports.

PR #53, opened 2022-11-10
Remove vite server components

christian-bromann:
As we are focusing on other feature sets now, let's remove these components for now and reduce extension size.


sourishkrout:
Sounds good. Let's maybe wait with the merge until we had a chance to ship `v0.3.0`. I will approve the PR once that has happened.

sourishkrout (approved):
Didn't go thru it file by file 👍 . Now that v0.3 shipped we can merge and break it in.

PR #57, opened 2022-11-15
Re-attempt to implement editing capabilities

christian-bromann:
My initial implementation to make cells editable was a bit flawed. This one should be more reliable, however still not perfect. Some observations:

- there is an edge case with new lines being added after new lines - could be because of VS Code trimming things or we do in Runme
- we have to keep original source and runme executable lines separated due keep any sorts of formatting
- which means we have to reload/reparse the markdown after pressing save

So for this to work we need to find a way to have VS Code re-serialize the content.


sourishkrout:
The new parser's notebook-based structure definitely makes it less finicky to implement a simple serializer. While this is an interesting start, I think it'll be difficult to deal with edge cases. Let's mark this branch as "draft". It's good input for the requirements gathering. Want to be sure it's not being merged yet, though.

sourishkrout (approved):
👍  - please note that we don't want to release the serialization impl as stable until we we had a chance to replace it with the new interface that includes de/-serialization which @adambabik is working on.

sourishkrout:
Closing in favor of https://github.com/stateful/vscode-runme/pull/81.

PR #59, opened 2022-11-15
Adding e2e tests

christian-bromann:
This test verifies if Runme Wasm is properly loaded by VS Code and an execution is triggered correctly.


.github/workflows/test.yml
```diff
@@ -13,7 +13,8 @@ jobs:
         with:
           node-version: 18
       - name: 📦 Install Dependencies
-        run: yarn
+        # use NPM here as Yarn as issues resolving WebdriverIO dependencies
+        run: npm install
```
christian-bromann, line 17: `@wdio/runner": "^8.0.0-alpha.558` depends on `"expect-webdriverio": "^4.0.0-alpha.3"` but it installed `v4.0.0` which was a [broken release](https://www.npmjs.com/package/expect-webdriverio/v/4.0.0) and I deprecated.

sourishkrout:
Reviewing now

.github/workflows/test.yml
sourishkrout: Perhaps we should just track a ticket to migrate from `yarn` to `npm` in dev too?

sourishkrout (approved):
Didn't work locally with `yarn` but once I installed deps with `npm` it ✅ .

PR #66, opened 2022-11-18
Support basic version control

christian-bromann:
This patch adds the ability to go back to previous versions of a document. As we start working on a solid foundation for storing changes to markdown files we want to ensure that users don't loose any data. This simple feature allows to store multiple versions of a document and load it back at any point in time.

Demo:
versioning

Possible improvements:
- rather than connecting the document to the local file system path we can also use Git information to make it even more flexible and allow reviving content made on other machines but on the same project in the same file


sourishkrout (changes requested):
Let's go ahead and put the implementation, as is, behind a setting (default turned off). In order to expose this to users the UX and some implementation details need to be completed. "Local version history" is a separate feature and overlaps with some of VS Code's core features (screenshot below). We can circle back to it a later.

As discussed, let's do the "your file is not version tracked" warning. It's a hedge to protect a likely small fraction of use-cases while editing features are experimental.

image

christian-bromann:
@sourishkrout I update the code to reflect the following behavior:

- if file is version controlled by git the save operation is executed based on implementation of #57
- if file is not version controlled and `runme.feature.versionHistory` setting is disabled, it throws an error message
- if file is not version controlled but `runme.feature.versionHistory` is enabled, save operation is executed

See example here:
Screenshot 2022-11-21 at 12 41 07

As version controlled I consider:
- workspace directory has a `.git` folder
- `git ls-files --error-unmatch ./path/to/file` is not failing

sourishkrout:
I'm getting this with an untracked `test.md`. Why not use the builtin git extension?

image

sourishkrout:
> > I'm getting this with an untracked `test.md`
> 
> As this file is not tracked into the repo, the error is expected. If our save mechanism fails, the user would loose their data.
> 

OK I think I'm following... but shouldn't the error message be the same as in your screenshot?

> > Why not use the builtin git extension?
> 
> IMO too much overhead for a simple check like this.

👌 

PS: Can we also remove the recommendation to change the setting from the message and say "please git track" it?

christian-bromann:
> PS: Can we also remove the recommendation to change the setting from the message and say "please git track" it?

What is the point of having the work around then if we mention it no where? Shall we remove it?

> but shouldn't the error message be the same as in your screenshot?

Good catch, fixed it.

sourishkrout:
> > PS: Can we also remove the recommendation to change the setting from the message and say "please git track" it?
> 
> What is the point of having the work around then if we mention it no where? Shall we remove it?

Between exposing as is or removing it, I think it's the latter.

PR #67, opened 2022-11-19
Allow to display images and other mime types

christian-bromann:
This patch allows to define a certain mime output type. Currently it will display an image if the mime type starts with `image/`. Others can be added. It makes our shell output cell component a bit more flexible.

__Demo:__

image

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -126,9 +126,15 @@ openssl rand -base64 32
 
 ## Non-Supported Languages
 
-These are shown as simple markdown, e.g:
+These are shown as simple markdowns, e.g:
 
 ```py { readonly=true }
 def hello():
     print("Hello World")
+```
+
+## Curl an image
+
+```sh { interactive=false, mimeType=image/png }
+curl https://lever-client-logos.s3.us-west-2.amazonaws.com/a8ff9b1f-f313-4632-b90f-1f7ae7ee807f-1638388150933.png 2>/dev/null
 ```
--- a/src/client/index.ts
+++ b/src/client/index.ts
@@ -17,21 +17,16 @@ export const activate: ActivationFunction = (context: RendererContext<void>) => 
   setContext(context)
   return {
     renderOutputItem(outputItem, element) {
-      const payload: CellOutput<OutputType> = outputItem.json()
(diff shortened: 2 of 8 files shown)


sourishkrout (changes requested):
1. The mime type in the attributes approach looks good to me 👍.
2. We should steer clear of reimplementing notebook renders for well-known mime types (most images, svg etc) since they are supported out of the box and if we want to have alternative ones, we should overload the existing renderers, as per the pluggable API reference. See demonstration snippet below.

(pick default renderers)
image

```
diff --git a/src/extension/executors/shell.ts b/src/extension/executors/shell.ts
index 49263dd..14240c0 100644
--- a/src/extension/executors/shell.ts
+++ b/src/extension/executors/shell.ts
@@ -34,13 +34,14 @@ async function shellExecutor(
    */
   function handleOutput(data: Buffer) {
     outputItems.push(data)
-    let item = NotebookCellOutputItem.json(<CellOutputPayload<OutputType.outputItems>>{
-      type: OutputType.outputItems,
-      output: {
-        content: Buffer.concat(outputItems).toString('base64'),
-        mime
-      }
-    }, OutputType.outputItems)
+    let item = new NotebookCellOutputItem(Buffer.concat(outputItems), mime)
+    // let item = NotebookCellOutputItem.json(<CellOutputPayload<OutputType.outputItems>>{
+    //   type: OutputType.outputItems,
+    //   output: {
+    //     content: Buffer.concat(outputItems).toString('base64'),
+    //     mime
+    //   }
+    // }, OutputType.outputItems)
 
     // hacky for now, maybe inheritence is a fitting pattern
     if (script.trim().endsWith('vercel')) {
```

christian-bromann:
@sourishkrout I removed the piece of code that tried to display known mime types, please have another look, thanks!

sourishkrout:
I think there used to be copy button on the output here:

image

christian-bromann:
> I think there used to be copy button on the output here:

Good catch, good we had e2e tests in place 😉

sourishkrout:
Who doesn’t love these e2e tests 😁

sourishkrout (approved):
👍

PR #75, opened 2022-12-05
Lower bundle size

sourishkrout:
Do we still have to include `node_modules` in packaging?

I believe in the past we kept the `node_modules` because of the web component stuff, no?


christian-bromann (approved):
👍

Issue #70, opened 2022-12-01
Implement Pseudoterminal to orchestrate execution


sourishkrout:
Proof reliable access to stdio and support for non-trivial TUX (character device and VT sequence stuff).

christian-bromann:
Closing as #77 was approved and merged.

PR #77, opened 2022-12-07
Use TaskProvider for Runme Execution

christian-bromann:
Implements #70

This patch implements a task provider for Runme related executions. It allows us to merge our previous `shell` and `task` executor into one and allow to stream stdout into the terminal and notebook. Furthermore does it allow us to populate markdown code blocks as tasks into VS Code, see demo below:

task
double


sourishkrout:
Just test drove this @christian-bromann. Good stuff! One thing I noticed is that TUX that involves cursor movements on that don't seem to work? E.g. when I do `vercel login` and it offers the different identity providers it won't prompt me for selection. Ideas?

sourishkrout:
It appears one can't re-run the same command multiple times... is that a known issue @christian-bromann?

christian-bromann:
> is that a known issue @christian-bromann?

Yes

sourishkrout:
> > is that a known issue @christian-bromann?
> 
> Yes

Is it a limitation we can overcome?

christian-bromann:
> Is it a limitation we can overcome?

I don't see why not. Looking into it.

christian-bromann:
@sourishkrout mind reviewing this PR? If we are happy with the new execution I can wrap this up writing unit / e2e tests for it.

sourishkrout (approved):
Overall LGTM 👍 . I like how you've used the VS Code Task APIs. Hopefully that will give Runme the robustness we're looking for. I managed to get into some edge cases, namely:
- When the neural net detects the correct lang but the CLI is unaware, the task is not being recognized (might get obsolete as we make changes to exec engine)
- It appears we've got an issue when stdin is being used that the terminal won't "complete" when no more reads are being made?
- I managed to get duplication (ever key stroke resulted in duplicate chars) within stdin, couldn't immediately reproduced it.

Notwithstanding any of these we should probably go ahead and merge since it's behind a feature flag for internal testing/hardening.

christian-bromann:
> * It appears we've got an issue when stdin is being used that the terminal won't "complete" when no more reads are being made?

Once you connect to the stdin socket the process can't quit. This will be also handled by the execution engine later on.

christian-bromann:
I think I found the issue: for some reasons using the same terminal for multiple task executions doesn't work well. I am not sure if this is a problem of VS Code or with our pseudo terminal:

bug

I created a new issue: https://github.com/stateful/vscode-runme/issues/113 and will merge this PR.

PR #80, opened 2022-12-08
Make sure dev won't accidentally bleed into releases

sourishkrout:
As we ramping up to get authoring/editing out the door we need a transitionary period where both exist. Want to me sure no accidents happen in the release pipeline etc.


adambabik (approved):
LGTM 👍 

During development, one typically will copy a WASM binary from https://github.com/stateful/runme into `./wasm` anyway.

PR #94, opened 2022-12-14
Bring back CLI button (and respective metadata)

sourishkrout:
Slightly refactored to better leverage types plus unit & e2e tests.

Relates to https://github.com/stateful/runme/issues/83. Tests will pass once PR (https://github.com/stateful/runme/pull/84) is merged.

PR #95, opened 2022-12-14
Some suggestions on PR #94

christian-bromann:


sourishkrout:
@christian-bromann using the implementation foregoes using the metadata type in a lot of places, no? Meaning we don't actually enforce it and it falls back to the very lenient `cell.metadata` vscode-interfaces?

christian-bromann:
Yeah, I guess this is just another option to do this:

- using a method that traverses the type
- extending VS Code native types

sourishkrout:
Makes sense. I'd say let's go with the original implementation. I could see a future refactor addressing this more generically. Another option, in my mind, is to define a sub-interface of `cell` that provides sufficient overlap with vscode's that's limit to only things that we use plus the metadata specificity. That feels like textbook separating impl from interface to me.

Issue #91, opened 2022-12-12
"New File" menu entry


sourishkrout:
Lets add a "New File" menu entry as per https://github.com/microsoft/vscode-github-issue-notebooks/blob/5ea44157dbd076dc6066540df431a17ca54010a8/package.json#L108-L113

_Originally posted by @sourishkrout in https://github.com/stateful/vscode-runme/issues/46#issuecomment-1320205889_

christian-bromann:
Moved into "Done" as referenced PR got merged.

PR #96, opened 2022-12-14
Add create new Runme notebook menu item

christian-bromann:
implements #91

create

### DIFF
--- a/__mocks__/vscode.ts
+++ b/__mocks__/vscode.ts
@@ -17,6 +17,7 @@ export const Uri = {
 }
 
 export const workspace = {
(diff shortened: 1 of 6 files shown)


sourishkrout:
@christian-bromann does it show up in the `File -> New File` options for you? It does not for me. The idea is not to run it directly from the command palette btw.

image

sourishkrout:
From the looks of it, the `"file/newFile"` is not in the right place in package.json and it likely needs an activation event.

sourishkrout (changes requested):
Please see comments.

christian-bromann:
@sourishkrout fixed contribution point and also added a link to the Runme docs:

newnew

sourishkrout (approved):
👍  with #100

PR #102, opened 2022-12-15
Allow user to lift save restriction

christian-bromann:
Demo:
lift


sourishkrout (approved):
👍 awesome. Let's merge!

PR #103, opened 2022-12-16
Prompt user for every cell when executing 'Run All'

### LINKED ISSUE #98: Ask for confirmation in between cells for "Run all"
Getting ahead of unintentional execution of cells as part of clicking "run all", let's consider inserting a confirmation step.

A later improvement could be automatically detecting "sensitive" cells and selectively asking for confirmation.

christian-bromann:
Fixes #98

runall

### DIFF
--- a/src/extension/kernel.ts
+++ b/src/extension/kernel.ts
@@ -1,4 +1,7 @@
-import { Disposable, notebooks, window, workspace, ExtensionContext, NotebookEditor, NotebookCell } from 'vscode'
+import {
+  Disposable, notebooks, window, workspace, ExtensionContext,
+  NotebookEditor, NotebookCell, NotebookCellKind
+} from 'vscode'
 
 import type { ClientMessage } from '../types'
 import { ClientMessages } from '../constants'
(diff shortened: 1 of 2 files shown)


sourishkrout:
I'd be curious to see if a modal-based notification vs quickpicker would work better here?

https://code.visualstudio.com/api/references/vscode-api#MessageOptions

christian-bromann:
Here is the modal based notification. A bit more subtle, folks might miss it depending on the window size.

parallel2

sourishkrout:
> Here is the modal based notification.

I thought notifs in modal mode would cover the viewport entirely. Did you set `modal: true` in the options? As per

image

christian-bromann:
Ah, gotcha, this would look like this:

parallel3

sourishkrout:
Thanks! Which one do you like best @christian-bromann? I'm torn between picker vs modal.

@admc any preference since you are usually opinionated about UI?

christian-bromann:
> Thanks! Which one do you like best @christian-bromann?

My order of liking is:

- __1st__ first solution, less dominant but visible enough and easy to navigate with the arrow keys
- __2nd__ last solution, very dominant, no selection possible with arrow keys but tab keys
- __3rd__ middle solution, can be overseen by the user, mouse handling required

admc:
- Quick picker
- Modal gets in the way of looking at the cell content
- Notif makes you worried something went wrong

sourishkrout:
Quick picker it is! Thanks for working thru the alternatives

PR #107, opened 2022-12-17
Use `runme` NPM package for dealing with WASM stuff

christian-bromann:
We have build an NPM package for deal with the whole WASM loading. We maintain types and exposed methods there.

Issues to resolve before merge:
- have `runme` export all types
- make it work with CJS (given that VS Code runs in that environment)


sourishkrout:
This would definitely make releasing more predicable and easier 👍.

sourishkrout:
Replacing WASM with GRPC.

PR #111, opened 2022-12-29
Set filename when running Runme through CLI

### LINKED ISSUE #109: CLI button does not appear to take filename into account
It does set `--chdir` but not `--filename`:

image

christian-bromann:
fixes #109


sourishkrout (approved):
👍 LGTM

PR #112, opened 2022-12-30
Don't show confirmation for notebooks with just one cell

### LINKED ISSUE #110: Run All confirmation shows for single cell markdowns
Probably a matter of checking if the cells array's size equals 1.

Repro

christian-bromann:
fixes #110


sourishkrout (approved):
👍 LGTM

PR #115, opened 2023-01-01
Basic annotations edit UX (actual UI needs more work)

sourishkrout:
Here's unfinished impl of how a custom renderer provides an intuitive UI (draft) to allow editing. Clean, intuitive, and in line with notebook UX does.

What's left:
- UI is rough and thrown together quickly. Needs refinement in separate PR. The horizontal only is not an optimal choice since viewport width is rarely wide enough.
- Currently behind feature flag (turn on `annotationsEdit` in vscode settings) so we can merge ahead of additional UI work in separate PR.

image

Closes https://github.com/stateful/vscode-runme/pull/116 (as obsolete).

ux


christian-bromann:
@sourishkrout interesting attempt, I haven't thought about using the `controller.createNotebookCellExecution(cell)` API to create an output. That said, my biggest concerns are the interplay between cell output and settings view and the limited UI space available. I am proposing an alternative that I mentioned in the original issue thread using a webview in the right sidebar in #116, what do you think?

sourishkrout:
> @sourishkrout interesting attempt, I haven't thought about using the `controller.createNotebookCellExecution(cell)` API to create an output. That said, my biggest concerns are the interplay between cell output and settings view and the limited UI space available. I am proposing an alternative that I mentioned in the original issue thread using a webview in the right sidebar in #116, what do you think?

Having this sidebar alternative in hand reaffirms the solution of this PR (#115). As mentioned, the strong association between (while being distinctly separate structurally) input & output cells is a core tenet of notebooks. Same goes for available space for UI (sidebar scores even lower here - uses up more space less efficiently), especially vertically which in the notebook world is basically infinite since one can always continue scrolling. Same goes for the "settings view" where overflow UI strategies with max-height scrollbars or tabs for subcategories. The interplay between input/output cells is at the core of the notebook UX concept.

Any UI comments you'd like to share @admc amongst the available three options (quick picker, cell output, sidebar)?

admc:
I can see the side-by-side webview being the right UX for something like the stackoverflow thread for an error message, or something that needs a bunch of space, but I don't think its right for these annotations. In this case I think we want the UX to be as close to the experience you get when editing raw markdown. So in short, lets tick with the inline cell editing experience thats closely tied to the rest of the cell functionality and doesn't require the users attention going to other places in the vs code UX.

sourishkrout:
Works end-to-end.

ux

src/extension/provider/annotations.ts
```diff
@@ -0,0 +1,64 @@
...
+import { getAnnotations } from '../utils'
+
+export class AnnotationsProvider implements NotebookCellStatusBarItemProvider {
+  constructor(private readonly kernel: Kernel) {
+    commands.registerCommand(
+      'runme.openCellAnnotations',
+      async (cell: NotebookCell) => {
+        try {
+          const exec = await kernel.createCellExecution(cell)
+          exec.start(Date.now())
```
degrammer, line 24: Can we encapsulate the start as part of createCellExecution, or there is any other use cases when we need more control in order to set the ellapsed time for cell execution?

src/extension/provider/annotations.ts
```diff
@@ -0,0 +1,64 @@
...
+              annotations: getAnnotations(cell),
+            },
+          }
+          await exec.replaceOutput([
+            new NotebookCellOutput([
+              NotebookCellOutputItem.json(json, OutputType.annotations),
+              NotebookCellOutputItem.json(json),
+            ]),
+          ])
+          exec.end(true)
```
degrammer, line 37: Do we need to use exec.end(true, Date.now()) instead?

src/extension/provider/annotations.ts
```diff
@@ -0,0 +1,64 @@
...
+          window.showErrorMessage(e.message)
+        }
+      }
+    )
+  }
+
+  async provideCellStatusBarItems(
+    cell: NotebookCell
+  ): Promise<NotebookCellStatusBarItem | undefined> {
+    if (!this.kernel.hasAnnotationsEditExperimentEnabled) {
```
degrammer, line 48: Question, if we run this flag check at the extension.ts level before registering the NoteBook provider, we can skip completely checking this flag each time a cell scrolls into view?

degrammer (commented):
Left some comments

src/client/components/annotations.ts
```diff
@@ -0,0 +1,157 @@
...
+    :host {
+      display: block;
+      font-family: Arial;
+    }
+
+    section {
+      padding: 10px;
+      border: 1px solid #444;
+      border-radius: 5px;
+      display: flex;
```
degrammer, line 26: I don't think this style is affecting at all the annotations container children, it's only applying a flex style to the grid container ((vscode-data-grid) parent of which children are being affected by a grid display. (Just in case that style is having any specific intention)

christian-bromann:
Suggestion on the UX: instead of having a checkbox and an input for mime type and name, just have a single input and only set the annotation if the input has a value. We could work with placeholders or labels to identify what the input stands for.

src/extension/provider/annotations.ts
sourishkrout: We don't since from an UX standpoint, I felt it's misleading to show an execution time on "sliding out a UI".

src/extension/provider/annotations.ts
sourishkrout: I see your point. I generally think this code needs a refactor. It feels like a bad code smell to me to rely on `kernel` being a closure here. However, this is the first time we're using a Notebook Output to provide a UI so I'd like to hold off with generalization until we have a better handle on it. However, your point about encapsulating impl detail here is well taken.

src/client/components/annotations.ts
sourishkrout: Would you mind taking care of this as part of working on the UI? I'm anything but a CSS ninja :-)

sourishkrout:
Merging behind feature flag so you can use this as baseline for UX/UI stuff @degrammer.

Issue #121, opened 2023-01-13
Hangs vscode.tasks.fetchTasks when there's no Readme.md in the workspace


biati-digital:
I created a simple extension like a year ago to display a list of tasks in the sidebar, everything has been working correctly until i installed the latest update of runme. Not sure if related to the TaskProvider you implemented recently.

![CleanShot 2023-01-13 at 09 38 10](https://user-images.githubusercontent.com/1219228/212359196-00eb73c6-a23a-498d-b5de-8b75c8c8272f.png)

in my extension i use the vscode API to fetch all tasks:

`let fetchedTasks: vscode.Task[] = await vscode.tasks.fetchTasks();`

If i enable your extension the above line never returns the task because your extension throws an error:

![CleanShot 2023-01-13 at 09 43 01](https://user-images.githubusercontent.com/1219228/212360070-d6b176d6-7d83-4e9c-b4ff-1b9fe8fc3bb3.png)

If i disable the runme extension `vscode.tasks.fetchTasks();` returns all tasks correctly.

biati-digital:
Even if i create an empty Readme file still does not work, disabling the readme extension seems to be the only solution, not sure what's happening.

sourishkrout:
Thanks for letting us know! Will get a fix out shortly @biati-digital.

sourishkrout:
Shipping a pre-release right now. Full release shortly after.

sourishkrout:
`0.4.1673629586` just hit the VS Code marketplace under the `pre-release channel`.

![image](https://user-images.githubusercontent.com/250527/212378806-ea801915-5216-41c3-9572-815249a4aa64.png)

biati-digital:
Thank you @sourishkrout i can confirm that the latest update fixes this error. Thank you and great work,

sourishkrout:
That's great! Thank you for confirming the fix.

PR #122, opened 2023-01-13
Handle file not found error

sourishkrout:
Fixes https://github.com/stateful/vscode-runme/issues/121. Thanks for reporting!


sourishkrout:
Merging and then working on adding a test case.

Issue #120, opened 2023-01-12
Sidebar panel (treeview) launcher


sourishkrout:
Let's build a vscode-native panel view that allows to launch one directly into a Runme notebook based on any markdown file (click on entry) available in the repository. The focus is on a flat list of files at first.

It might make sense to provide custom sorting where any file starting with capitalized letters shows up higher in the list, e.g. `README.md`, `CONTRIBUTING.md`, `DEV.md` over 'my-markdown-docs.md`. However, let's iterate on the UX details once a baseline version is available.

See "NPM Scripts" as reference:
![image](https://user-images.githubusercontent.com/250527/212097456-a2a525f4-ad36-442b-a425-bbf7171dddfb.png)

PR #123, opened 2023-01-13
Runme notebooks launcher

degrammer:
First iteration for having a markdown filter tree as part of runme extension.
TODO:
- [ ] Make folder exclusion configurable (e.g exclude node_modules folder)
- [x] Open markdown files as runme notebooks
- [ ] Make tree navigation more user friendly, reflecting the original folders structure from the workspace
- [ ] Allow right click over specific folders to filter markdown files

Fixes [https://github.com/stateful/vscode-runme/issues/120](https://github.com/stateful/vscode-runme/issues/120)

https://user-images.githubusercontent.com/4001529/212779676-b732bca5-bcf5-412f-83eb-df948edb544e.mp4


sourishkrout:
Looks like a great start 👍

sourishkrout:
> * [ ]  Make folder exclusion configurable (e.g exclude node_modules folder)
It'd be great if that was a simple toggle like the timeline panel offers
image

> * [ ]  Open markdown files as runme notebooks
Yep, that's a must have

> * [ ]  Make tree navigation more user friendly, reflecting the original folders structure from the workspace
Let's definitely experiment with that. Currently the UX is a little rocky if entries aren't uncollapsed by default.

> * [ ]  Allow right click over specific folders to filter markdown files
I'd put this in the nice to have category.

Outside of the todos, I'd like to do some editing on panel name etc closer to the finish line.

src/extension/provider/launcher.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    return Promise.resolve(folderMarkdownItems)
+  }
+
+  refresh(): void {
+    this._onDidChangeTreeData.fire()
+  }
+
+  getRunmeFilesFromWorkspace(onComplete: (value: RunmeFile[]) => void): any {
+    glob('**/*.md', { cwd: this.workspaceRoot, ignore: 'node_modules/**', absolute: true }, (err, files) => {
```
sourishkrout, line 90: @degrammer you do want to use the VS Code API's way to glob/traverse the FS. The reason is that vscode can run against virtual filesystems when attached to a SSH host or directly attached to a Github Repository. It's really only guaranteed that local FS is available when VS Code is run as Desktop app. It's something to get used to building inside of VS Code.

image
Can't deep link to the API hence screenshot: https://code.visualstudio.com/api/references/vscode-api#workspace

src/extension/provider/launcher.ts
```diff
@@ -0,0 +1,118 @@
+import { join, basename, dirname, resolve } from 'node:path'
+
+import * as vscode from 'vscode'
```
christian-bromann, line 3:
```suggestion
import vscode from 'vscode'
```

tests/extension/provider/launcher.test.ts
```diff
@@ -0,0 +1,40 @@
+import { vi, describe, it, expect } from 'vitest';
```
sourishkrout, line 1: Nit: Looks like this file need a eslint fix-all run. Based on the current rules I won't pass linting.

package.json
```diff
@@ -263,6 +263,20 @@
           }
         }
       }
+    ],
+    "views": {
+      "explorer": [
+        {
+          "id": "runme.launcher",
+          "name": "Runme Notebooks"
```
sourishkrout, line 271: I believe there's an attribute that controls default visibility (collapsed/uncollapsed). We probably want to decide how front and center we want to be?

package.json
```diff
@@ -275,6 +289,7 @@
     "postinstall": "npm run download:wasm",
     "test": "run-s test:*",
     "test:lint": "eslint src --ext ts",
+    "test:lint:fix": "eslint src --ext ts --fix",
```
sourishkrout, line 292: This won't catch the `tests` directory :-D

sourishkrout (changes requested):
Looks fantastic 👏 ! My and Christian's comments are just nits. Should be easily fixed and ready for merge 👍  after (no additional review required).

In future enhancements (just writing down here so we don't forget):
- Consider adding the treeview highlighting the currently active notebook file
- Action buttons to quickly collapse/uncollapse all entries analogous to the file explorer.

Let's port these items plus the open check boxes above into a new issue for future feature consideration.

package.json
degrammer: Added a default collapsed visibility

package.json
degrammer: Right, fixed :)

tests/extension/provider/launcher.test.ts
degrammer: Fixed

src/extension/provider/launcher.ts
degrammer: Fixed, thanks

PR #114, opened 2022-12-30
Refactor output buttons

christian-bromann:
Implements: #99

Demo:
refactor

There are some outstanding questions to resolve (see original issue). Once these are resolved and this solution is approved I will add/fix unit tests and move PR out of draft

PR #125, opened 2023-01-15
Remove stale open terminal button

### LINKED ISSUE #124: Open Terminal can become stale
After creating a terminal for a background process, the button/PID will linger even after the terminal is destroyed:

mxsdev:
Fixes #124 

This may conflict with #114, but it shouldn't be hard to integrate this if/when that PR gets approved 👍


sourishkrout:
Thanks so much for the PR contribution @mxsdev!

Will do a review in the next couple of days.

src/extension/utils.ts
```diff
@@ -171,3 +171,50 @@ export async function initWasm(wasmUri: Uri) {
     }
   )
 }
+
+// eslint-disable-next-line max-len
+// adapted from https://github.com/microsoft/vscode/blob/5e4c5c78122d02469ba19d778588ee05fe841f5a/extensions/markdown-language-features/server/src/util/dispose.ts#L41
+export abstract class DisposableRegistrar implements vscode.Disposable {
```
sourishkrout, line 177: It does make sense to me wanting to avoid DRY for classes implementing `vscode.Disposable` to clean up after themselves. However, I think it'd rather move this to an issue to consider brought adoption of the pattern across the codebase. Not saying this is a bad idea, rather consider a consolidated effort to standardize on it than introducing something in a single one place in the code base as part of another effort.

sourishkrout (changes requested):
Overall this is a great PR and I'd love to remove the stale buttons. It likely will take some time for the output buttons to land and meanwhile this is a great UX improvement.

Please see my comment regarding the `vscode.Disposable` mixin. I'd rather split that out and have a discussion with the other devs how to adopt it across the codebase.

Also, sorry, we merged a PR early that is causing conflicts. Happy to help with resolving those if it's too complicated.

src/extension/utils.ts
mxsdev: Gotcha! I'll move the code from `DisposableRegistrar` into `ShowTerminalProvider` for now, and go ahead and make an issue about wider adoption of the `Disposable` pattern.

sourishkrout:
Thanks for the changes @mxsdev. Just noticed that it seems to swallow a background task's `Open Terminal` button even when the task is still active.

Example: 3rd code block from the top: https://github.com/stateful/vscode-runme/blob/main/examples/README.md

mxsdev:
@sourishkrout 

Hey, I took a look but was unable to reproduce the issue. Is the video below not intended behavior? Or are you referring to the initial period of time when the command is running and there is no open terminal button? (If so to the latter, that, I think, is unrelated to this PR) 

<p align="center">

</p>

sourishkrout:
> Hey, I took a look but was unable to reproduce the issue. Is the video below not intended behavior? Or are you referring to the initial period of time when the command is running and there is no open terminal button? (If so to the latter, that, I think, is unrelated to this PR)

Entirely possible that I got this wrong. Retesting right now. Will report back.

sourishkrout (approved):
Works as expected 👍 . Working on resolving the conflicts (since I'm the one who caused them).

sourishkrout:
Test are hanging. Something isn't right there. Working on it.

sourishkrout:
Might want to check out: https://github.com/stateful/vscode-runme/pull/125/commits/645e9a04a9b262d4153aee0f27d2292b5a45b2bb @mxsdev.

Specifically https://github.com/stateful/vscode-runme/pull/125/commits/645e9a04a9b262d4153aee0f27d2292b5a45b2bb#diff-c2dae4fd6416959b0a0f5c7e8c9b1065fb22fc502e149cd580fe97ef4686fc19L6-L12 seemed to have caused an infinite loop of sorts.

Ever since the base had changed there was a "different" mocking implementation that clashed which I had to generalize. Figured it'd be of advantage to be able to share mocks.

mxsdev:
@sourishkrout 

Ah, I see. Big thanks for clearing that up - I probably would've struggled to figure that out!

sourishkrout:
> @sourishkrout
> 
> Ah, I see. Big thanks for clearing that up - I probably would've struggled to figure that out!

So did I 😆!

PR #126, opened 2023-01-18
GRPC serializer behind feature flag

sourishkrout:
De-/serialization works as long as `runme server --address /tmp/runme.sock` (address is hardcoded for now) is running.

Todos:
- [x] Refactor to allow WASM and GRPC to coexist
- [x] Feature flag `runme.experiments.grpcSerializer`
- [x] Fix cell metadata changes fail serialization
- [x] Fix remaining issues GRPC interface (mangles frontmatter, `langId` vs `languageId`, etc)
- [x] Skipped unit tests, need to be covered as e2e when server lifecycle's figured out
- [ ] Figure out server lifecycle & distribution (separate ticket)


sourishkrout:
@adambabik if you want to try it out yourself. The extension won't launch the server for you (yet).

adambabik:
Deserializing works but I can't save an edited file:

```
Failed to save 'README.md': 13 INTERNAL: Request message serialization failure: 
```

It seems that the request does not reach `runme` at all.

sourishkrout:
> Deserializing works but I can't save an edited file:
> 
> ```
> Failed to save 'README.md': 13 INTERNAL: Request message serialization failure: 
> ```
> 
> It seems that the request does not reach `runme` at all.

This looks more like a proto serialization mismatch to me. Lemme test with the fixed languageId inconsistency.

sourishkrout:
> > Deserializing works but I can't save an edited file:
> > ```
> > Failed to save 'README.md': 13 INTERNAL: Request message serialization failure: 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It seems that the request does not reach `runme` at all.
> 
> This looks more like a proto serialization mismatch to me. Lemme test with the fixed languageId inconsistency.

Confirming. Seeing the same issue. Lemme do some digging.

sourishkrout:
Looks like the Buf registry is not up to date with the proto change @adambabik. Did you push?

sourishkrout:
The serialization problem is fixed now 👍 . FYI, going to rebase this branch to resolve conflicts.

christian-bromann:
I checked on the unit test issue and wasn't really able to find a esbuild/vite configuration that would allow to resolve the problem. I can think of two options now:

- mock out this dependency (which I would recommend)
- create a setup script that would add that extension

What do you think works best?

sourishkrout:
> * mock out this dependency (which I would recommend)

Thanks for sifting through vite(st) config stuff @christian-bromann. I agree, mocking is the best option since the logic here is rather thin and the integration with the GRPC servers is better covered in e2e tests. I'll take that route.

sourishkrout:
Merging this now to prevent long-running rebasing/conflict resolution. Let's do a retro-review when you're back online please @degrammer mostly so you get a sense where this is moving. Happy to go over it in real-time.

PR #135, opened 2023-01-20
Copy changes and slight UI adjustment

sourishkrout:
I still would like the "?" at some point where we can describe *why* now that the UI show the "what". I'd be comfortable shipping as is though.

Will squash upon merge.


degrammer (approved):
LGTM

PR #138, opened 2023-01-23
Add .gitignore files/folder to exclude list from the launcher

degrammer:


sourishkrout:
The `.gitignore` seems to work well now. However, I'm still getting absolute paths:

image

It'd expect folder entries to be trimmed all the way up to `.../stateful.com`.

sourishkrout (changes requested):
See comment please

sourishkrout (approved):
👍 LGTM - thanks for adding the annotations tests

PR #141, opened 2023-01-24
Auto update project dependencies

christian-bromann:
This patch adds two files (`dependabot.yml` and `update.yml`) that allows us to have dependencies automatically updated (once every week), if our pipeline passes. I've been using this approach in WebdriverIO now very successfully and thought we could adapt this too.


sourishkrout (approved):
👍

Issue #132, opened 2023-01-19
Add husky pre-push hook


degrammer:
Add a husky pre-push hook that prevents pushing if the branch has one of the following:
- Failed tests
- Linting errors

christian-bromann:
@degrammer please have a look at my PR. I left it with just checking Eslint because I feel like adding unit test to the pre commit hook could slow down ones workflow and is sometimes not needed, e.g. if there are changes to the readme or other parts of the code base that aren't impacted by unit tests. That said, I think we can be always smarter what we want to do and this could be a good start.

christian-bromann:
Merged, so closing.

PR #142, opened 2023-01-24
Add husky pre-commit and pre-push hooks

christian-bromann:
implements #132


sourishkrout:
Doing the review since @degrammer is out for a few days.

sourishkrout (approved):
Works for me locally 👍

PR #146, opened 2023-01-25
Mimic POSIX var substitution behavior

sourishkrout:
This will change the behavior of using `$(...)` expressions inside of exported ENV vars `export MY_VAR=$(echo 'test')`.

Old behavior inside of Runme notebook; notice gap (it's a `\n` but displayed as space by the markdown renderer) inside of MY_VAR:
```sh
$ export MY_CWD="$(pwd)"
$ export MY_VAR="$MY_CWD:$PATH"
$ echo $MY_VAR
/Users/sourishkrout/Projects/stateful/oss/runme :/Users/sourishkrout/.wasmtime/bin:/opt/homebrew/opt/libpq/bin:/Users/sourishkrout/go/bin:/Users/sourishkrout/.asdf/shims:/opt/homebrew/opt/asdf/libexec/bin:/Users/sourishkrout/bin:/opt/homebrew/bin:/opt/homebrew/sbin:/usr/local/bin:/Library/Apple/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Users/sourishkrout/.cache/zsh4humans/v5/fzf/bin
```

New behavior; no special endings (e.g. '\n'):
```sh
$ export MY_CWD="$(pwd)"
$ export MY_VAR="$MY_CWD:$PATH"
$ echo $MY_VAR
/Users/sourishkrout/Projects/stateful/oss/runme:/Users/sourishkrout/.wasmtime/bin:/opt/homebrew/opt/libpq/bin:/Users/sourishkrout/go/bin:/Users/sourishkrout/.asdf/shims:/opt/homebrew/opt/asdf/libexec/bin:/Users/sourishkrout/bin:/opt/homebrew/bin:/opt/homebrew/sbin:/usr/local/bin:/Library/Apple/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Users/sourishkrout/.cache/zsh4humans/v5/fzf/bin
```

Some tools are more tolerant than others. E.g. the PATH is sensitive and encoded strings passed into tools like `grpcurl` will fail as a result. Others are more fault-tolerant where they likely sanitize/trim previously extrapolated vars taken as input.

Whether or not this is exactly mimicking POSIX behavior it's far "less wrong" than the old behavior. Leaned on explanation from https://unix.stackexchange.com/questions/164508/why-do-newline-characters-get-lost-when-using-command-substitution/164548#164548 and specifically https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_03

This PR will also fix a bug where future var substitutions would not retain previously evaluated vars.

### DIFF
--- a/src/extension/executors/utils.ts
+++ b/src/extension/executors/utils.ts
@@ -62,6 +62,7 @@ export async function retrieveShellCommand (exec: NotebookCellExecution) {
       const expression = placeHolder.slice(2, -1)
       const expressionProcess = cp.spawn(expression, {
         cwd,
+        env: {...process.env, ...stateEnv },
         shell: true
       })
       const [isError, data] = await new Promise<[number, string]>((resolve) => {
@@ -69,6 +70,7 @@ export async function retrieveShellCommand (exec: NotebookCellExecution) {
         expressionProcess.stdout.on('data', (payload) => { data += payload.toString() })
         expressionProcess.stderr.on('data', (payload) => { data += payload.toString() })
         expressionProcess.on('close', (code) => {
+          data = data.trim()
           if (code && code > 0) {
             return resolve([code, data])
           }
--- a/tests/extension/executors/utils.test.ts
+++ b/tests/extension/executors/utils.test.ts
@@ -124,6 +124,34 @@ test('can support expressions', async () => {
   expect(ENV_STORE.get('foo')).toBe('foobar')
 })
 
+test('can recall previous vars', async () => {
+  const exec1: any = {
+    cell: {
+      metadata: { executeableCode: 'export foo=$(echo "foo$BAR")' },
+      document: {
+        getText: vi.fn().mockReturnValue('export foo=$(echo "foo$BAR")'),
+        uri: { fsPath: __dirname }
+      }
+    }
+  }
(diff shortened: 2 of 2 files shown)


adambabik:
@sourishkrout it makes sense. It seems like the old behaviour has always been adding a new line at the end. What's the cause of that? 

`export MY_CWD="$(pwd)"` should not end with a new line as `$(pwd)` does not contain it.

sourishkrout:
> @sourishkrout it makes sense. It seems like the old behaviour has always been adding a new line at the end. What's the cause of that?
> 
> `export MY_CWD="$(pwd)"` should not end with a new line as `$(pwd)` does not contain it.

As discussed out of band: The `\n` was introduced since the notebook execution currently spawns a subshell for every `$(...)` executing it as `...` and retaining it in an ENV store map in memory. In essence, it's actually not doing shell based var substitution. The fix is naive and will just trim wrapping whitespace (probably not 100% POSIX compliant but far less wrong) until we move all of this into the GRPC KernelService.

Anyways, just summarizing for the benefit of everybody else.

adambabik (approved):

PR #179, opened 2023-02-06
Baseline modeling of annotations to allow data validation

sourishkrout:
Unfinished. What's left:

- [x] Validate serialization and handle errors appropriate
- [x] Form-level validation in Annotations UI
- [x] Testing

Error handling mechanism:
The original PR throws an error when finding invalid values. The latest version introduces safe defaults so that the Notebook can be opened (it fallbacks to the annotation's default values). The idea is that the user can still run the code block while seeing an error message that the configuration values are using the wrong data. This gives the user to fix the issue and save it. We are also displaying validation messages on the editing experience.

Screenshot 2023-02-14 at 12 57 11 PM
Screenshot 2023-02-14 at 12 58 54 PM

https://user-images.githubusercontent.com/4001529/218818809-3efca890-6b9b-4ec7-bf45-1e078c3b3c2f.mp4

### DIFF
--- a/src/client/components/annotations.ts
+++ b/src/client/components/annotations.ts
@@ -1,10 +1,10 @@
 import { LitElement, css, html } from 'lit'
-import { customElement, property, state } from 'lit/decorators.js'
+import { customElement, property } from 'lit/decorators.js'
 import { when } from 'lit/directives/when.js'
 
 import { ClientMessages } from '../../constants'
 import type { ClientMessage, CellAnnotations, CellAnnotationsErrorResult } from '../../types'
-import { getContext, tryBoolean } from '../utils'
+import { getContext } from '../utils'
 import { CellAnnotationsSchema, AnnotationSchema } from '../../schema'
 import '@vscode/webview-ui-toolkit/dist/data-grid/index'
 
@@ -99,9 +99,6 @@ export class Annotations extends LitElement {
   @property({ type: Object, reflect: true })
   validationErrors?: CellAnnotationsErrorResult
 
-  @state()
-  private shouldRender: boolean = false
-
   #desc(id: string): string {
     return this.#descriptions.get(id) || id
   }
@@ -123,8 +120,7 @@ export class Annotations extends LitElement {
 
     const propVal: any = { 'runme.dev/uuid': this.annotations['runme.dev/uuid'] }
     const propName = e.target.id
-    const value = this.#getTargetValue(e)
-    const targetValue = tryBoolean(value)
+    const targetValue = this.#getTargetValue(e)
 
     const parseResult = CellAnnotationsSchema.safeParse({
       [propName]: targetValue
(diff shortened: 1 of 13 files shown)


sourishkrout:
@degrammer: please take a look. I'd like us to use zod for any data interaction with the user. This PR is a prototypical implementation. It might be wise to separate the `getAnnotations` function from the validation and only use the latter in the web-component that exposes the annotations form. Please make recommendations how you'd like to complete this implementation.

src/types.ts
```diff
@@ -104,11 +105,29 @@ export interface RunmeTaskDefinition {
...
-}
+const falseyBoolean = z.preprocess((subject) => {
+  if (typeof subject === 'string' && subject.toLowerCase() === 'false') {
+    return false
+  }
+  return Boolean(subject)
+}, z.boolean())
+
+export const CellAnnotationsSchema = z.object({
+  'runme.dev/uuid': z.string().optional(),
```
christian-bromann, line 116: I'ld suggest to move this into `constants.ts` as this is rather an object initiation than a type definition

sourishkrout: Totally. Or if we'll even have more schemas we could just move them to a `schemas.ts`.

degrammer: Moved all the zod schemas to schemas.ts

src/client/components/annotations.ts
```diff
@@ -69,27 +94,68 @@ export class Annotations extends LitElement {
...
+        this.validationErrors.errors[propName] = fieldErrors[propName]
+      }
+      // Re-render the form
+      return this.shouldRender = !this.shouldRender
+    }
+
+    if (this.validationErrors?.errors && this.validationErrors.errors[propName]) {
+      delete this.validationErrors.errors[propName]
+      // Re-render the form
+      this.shouldRender = !this.shouldRender
```
degrammer, line 148: I don't like how I'm mutating this object here, suggestions are welcome.

sourishkrout:
Any chance you could rebase this against `main` please @degrammer? I fixed some issues with cell associations when editing annotations. Just to be sure I'm testing from a known good state.

degrammer:
> Any chance you could rebase this against `main` please @degrammer? I fixed some issues with cell associations when editing annotations. Just to be sure I'm testing from a known good state.

Done, rebase against main

sourishkrout:
I believe my initial schema had the now `bools` modeled as strings because that's what the parser expects:

image

@degrammer does serialization/deserialization work for you? it does not work for me

degrammer:
> I believe my initial schema had the now `bools` modeled as strings because that's what the parser expects:
> 
> image
> 
> @degrammer does serialization/deserialization work for you? it does not work for me
> 
> There's also a chance that the code converted the bools to string `onChange` to conform with the parser. Either way, that seems to currently not work unless I'm overlooking something.

It works for me, probably I don't have some settings you have, let's discuss via Zoom

src/client/components/annotations.ts
sourishkrout: for re-render please try `this.requestUpdate()`

degrammer:
Fixed the issue: Screenshot 2023-02-15 at 2 27 50 PM

sourishkrout:
Worked. Good to merge @degrammer 👍 . Let's do a edge/pre-release on it right after so we can dogfood it for a few days.

PR #187, opened 2023-02-09
Fix url parsing for git@ urls

christian-bromann:
This broke with the last release. I am working on e2e tests but feel free to review

### DIFF
--- a/tests/extension/handler/utils.test.ts
+++ b/tests/extension/handler/utils.test.ts
@@ -4,10 +4,9 @@ import { describe, it, expect, vi, beforeEach } from 'vitest'
 import { Uri, workspace, window } from 'vscode'
 
 import {
-    getProjectDir, getTargetDirName, getSuggestedProjectName
+    getProjectDir, getTargetDirName, getSuggestedProjectName, parseParams
 } from '../../../src/extension/handler/utils'
 
-vi.mocked(Uri.parse).mockImplementation((input: any) => input)
 vi.mocked(Uri.joinPath).mockImplementation(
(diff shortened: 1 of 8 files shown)


tests/extension/handler/utils.test.ts
```diff
@@ -87,3 +85,13 @@ describe('getSuggestedProjectName', () => {
         expect(window.showErrorMessage).toBeCalledTimes(1)
     })
 })
+
+describe('parseParams', () => {
+  it('should parse params to be used safely', () => {
+    const usp = new URLSearchParams('fileToOpen=foo;bar loo&repository=git@github.com/org/project;foo bar.git')
+    expect(parseParams(usp)).toEqual({
+      fileToOpen: 'foo%3Bbar%20loo',
+      repository: 'git@github.com/org/project%3Bfoo%20bar.git'
```
sourishkrout, line 94: @christian-bromann is a URL encoded repository ref not an invalid ref?

sourishkrout (changes requested):

tests/extension/handler/utils.test.ts
christian-bromann: I am not sure what you mean by that. This would fail later because the repo is not found this way, however it prevents injecting malicious stuff into the terminal.

christian-bromann:
@sourishkrout fyi: I wasn't able to set up a e2e scenario as it seems impossible to overcome the alerts of starting VS Code from Chrome 😞

sourishkrout (approved):

PR #201, opened 2023-02-13
Code cells always require session UUIDs

### LINKED ISSUE #188: Configuration changes for code cells don't persist
To reproduce: create a new markdown file and try to modify the configs. It seems that disabling interactive mode doesn't persist, e.g.:

bug-authoring

sourishkrout:
Fixes #188. Please review @degrammer or @mxsdev whoever gets to it sooner.

### DIFF
--- a/src/extension/serializer.ts
+++ b/src/extension/serializer.ts
@@ -49,6 +49,10 @@ export abstract class SerializerBase implements NotebookSerializer, Disposable {
     this.disposables.forEach(d => d.dispose())
   }
 
+
+  /**
+   * Handle newly added cells (live edits) to have UUIDs
+   */
   protected handleNotebookChanged(changes: NotebookDocumentChangeEvent) {
     changes.contentChanges.forEach((contentChanges) => {
       contentChanges.addedCells.forEach((cellAdded) => {
@@ -213,7 +217,7 @@ export abstract class SerializerBase implements NotebookSerializer, Disposable {
       }
 
       if (cell.kind === NotebookCellKind.Code) {
-        // todo(sebastian): decide if the serializer should own lifecycle
+        // serializer owns lifecycle due live edits bypass deserialization
         cell.metadata = SerializerBase.addCellUuid(elem.metadata)
       }
 
--- a/src/extension/serializer.ts
+++ b/src/extension/serializer.ts
@@ -8,6 +8,10 @@ import {
   NotebookCellData,
   NotebookCellKind,
   CancellationToken,
+  workspace,
+  WorkspaceEdit,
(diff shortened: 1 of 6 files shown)


src/extension/serializer.ts
```diff
@@ -165,9 +216,10 @@ export abstract class SerializerBase implements NotebookSerializer {
         )
       }
 
-      cell.metadata = { ...elem.metadata }
-      // todo(sebastian): decide if the serializer should own lifecycle
-      cell.metadata['runme.dev/uuid'] = uuidv4()
+      if (cell.kind === NotebookCellKind.Code) {
+        // serializer owns lifecycle due live edits bypass deserialization
```
mxsdev, line 220:
```suggestion
        // serializer owns lifecycle because live edits bypass deserialization
```

mxsdev (approved):
Worked through the problem and your code looks good to me! One thing I noticed though is the following dubious on line 107 of `kernel.ts`:

```ts
          if (cell.metadata?.['runme.dev/uuid'] === payload.output.annotations['runme.dev/uuid']) {
            editCell = cell
            break
          }
```

In the current version of `runme`, since newly added cells have an undefined `uuid`, this can actually result in editing one cell affecting a different cell (the first added cell with no `uuid` field) 😬

I think we should either add a `cell.metadata?.['runme.dev/uuid'] !== undefined` check here, and/or log an error in this case, because we may run into another error sometime down the line caused by an undefined `uuid` field (for instance if we add new ways to add cells or serialize notebooks).

src/extension/serializer.ts
```diff
@@ -26,12 +31,58 @@ const DEFAULT_LANG_ID = 'text'
...
+    changes.contentChanges.forEach((contentChanges) => {
+      contentChanges.addedCells.forEach((cellAdded) => {
+        if (
+          cellAdded.kind !== NotebookCellKind.Code ||
+          cellAdded.metadata['runme.dev/uuid'] !== undefined
+        ) {
+          return
+        }
+
+        const notebookEdit = NotebookEdit.updateCellMetadata(
```
mxsdev, line 66: Oh yeah, one more thing: We should make sure this `NotebookEdit` is not `Ctrl+z`'able, since that would make it possible to "undo" the operation of adding a uuid.

sourishkrout: It is not. I did look into this initial when I first started using this API.

sourishkrout:
> May even be worth writing lines in our tests to ensure `uuid` is never undefined.

I will create a follow up ticket to add tests. Want to get the fix out the door asap.

Issue #202, opened 2023-02-13
gRPC Runner Service integration


mxsdev:
We want to integrate the newly completed [gRPC interface](https://github.com/stateful/runme/pull/119)

 - Implement some kind of interface around gRPC generated files (probably in `extension/runner.ts`)
 - Bring runner interface into `kernel.ts`, specifically `_doExecuteCell` (favoring simple/naive implementation)
   - Probably separate from `executors` system, which we're thinking about moving away from
 - Not concerned about bundling executable, gRPC transport, dealing with generated files for now
 - Tests not a priority
 - Handle PTY if it seems reasonable

PR #134, opened 2023-01-20
Annotations UI Improvements

degrammer:
Introducing UI tweaks for annotations, includes:
- placeholders instead of labels for textinput
- Remove grid component for composing the layout, it adapts to the screen width and behaves well under zoom in/out

TODO: Include an explanatory "?" for each element

demo-ui

dem

Solves [https://github.com/stateful/vscode-runme/issues/131](url)

Light Mode

light

Dark Mode
dark

Issue #131, opened 2023-01-19
Annotations UI Improvements


degrammer:
The current annotations UI needs improvements, review the layout display (probably a vertical one) since the current distribution may need more screen real state, it can easily run out of space for certain screens or even when zooming.

![Image](https://user-images.githubusercontent.com/4001529/213330107-e2bec4c3-e0a4-41c5-8f92-76ef0b87a047.png)

christian-bromann:
To re-iterate some comments I made before in other threads:
- I would merge the inputs for mime type and name to be a single input element which value is omitted if empty
- as mentioned in #116 the current solution to integrate these inputs right below the code block limits the amount of screen real estate we can use, it works with the current 5 inputs but will get out of hand when that increases, I personally still feel that a sidebar provides us more options

sourishkrout:
As discussed, we want to make a best effort iteration of the UI that's there. This feature will ship tomorrow. As per discussion on the closed sidebar approach PR (https://github.com/stateful/vscode-runme/pull/115), it's been decided to move ahead.

@degrammer one thing that's somewhat important to me is to have a "?" on every annotation that ideally shows a short description of its effect on the cell execution.

degrammer:
Working on the "?" progress for the enhancements here https://github.com/stateful/vscode-runme/pull/134

sourishkrout:
Done.

Issue #130, opened 2023-01-19
Runme notebooks panel:  Add action buttons to collapse/uncollapse notebooks


degrammer:
Add an action button to runme notebooks panel to quickly collapse / uncollapse folders.

sourishkrout:
Implemented by https://github.com/stateful/vscode-runme/pull/145.

PR #205, opened 2023-02-14
Integrate gRPC Runner Service

mxsdev:
Solves #202 

Currently facing the following issues:

@adambabik 

 - `InputData` sent in the `sendRawInput` function (located in `runner.ts > GrpcRunnerProgramSession`) is completely empty by the time it gets to `runnerService.Execute` on the server. It's getting totally dropped, and this is a big blocker for interactive support. Do you have any idea why this might be? @sourishkrout please test and let me know if you have the same issue/any thoughts here, I have absolutely no idea what could be causing this...
 - I get the following trace when running Execute. I think what's happening here is that the program has exited, so the stream has ended, but `srv.Recv()` still gets called one last time, resulting in this "rpc Canceled" error. If this is not a client-side issue, I think we should make an issue about it on the `runme` repo.
```
2023-02-14T15:23:11.950-0600	INFO	runner/service.go:122	running Execute in runnerService
2023-02-14T15:23:11.969-0600	INFO	runner/service.go:204	finished command execution	{"exitCode": 0}
2023-02-14T15:23:11.969-0600	INFO	runner/service.go:173	failed to receive requests; stopping the command	{"error": "rpc error: code = Canceled desc = context canceled"}
2023-02-14T15:23:11.969-0600	INFO	runner/service.go:175	failed to stop the command	{"error": "os: process already finished", "errorVerbose": "os: process already finished\ngithub.com/stateful/runme/internal/runner.(*command).Stop\n\t/Users/mxs/Projects/runme/internal/runner/command.go:343\ngithub.com/stateful/runme/internal/runner.(*runnerService).Execute.func1\n\t/Users/mxs/Projects/runme/internal/runner/service.go:174\nruntime.goexit\n\t/opt/homebrew/Cellar/go/1.19.5/libexec/src/runtime/asm_arm64.s:1172"}
```
 - Is there any way to request a program exit other than sending an interrupt? Currently if the user requests a command stop,  the client sends an `EOF`, but then the program continues running. I think maybe that the program should try to stop if its still running during an `EOF`, since an interrupt can be cancelled by the program, but something like `cmd.Stop()` cannot - what do you think?

@sourishkrout 

 - I added the main functionality to `kernel.ts` as an inline async function. Much of this is copy/pasted and/or adapted from `taskExecutor` and `shellExecutor`. I want to bring this to its own file/method, but its not clear exactly how. I basically see two ways to proceed:
   1. Add a new `executor` function which runs the given runner. This is basically what I've done already, but results in lots of duplication/furthering the aforementioned issue of cementing the `executor` model.
   2. Create an abstraction - my idea is something like an abstract class `ExecutionNode`, so for the runner it will be `RunnerExecutionNode`. For now it will be implemented rather simply (maybe it flatly produces some kind of "`ExecutionResult`" interface), but I think this is a good step towards things like multiple output stages, cleaner input utilization, chaining, etc. 
 - Basically, if I understood correctly, in our 1:1 you were describing a desire for a kind of "middle ground" solution between a simplistic function (executor) and a full-blown solution. I think I can implement a class (`ExecutionNode`) which satisfies this desire and also can be built upon going forward. Maybe the exact abstraction described in (ii) isn't the one you want to go with, but in any case I don't think it will be a significant development effort to implement it. But I need your approval before going this way, as we could easily just do (i) and then work towards (ii) later/in follow up PRs, and I know this feature is time-sensitive.

Other issues:

 - Currently there is one environment per `Kernel`. We need to set-up a system that puts the session-id on the notebook metadata if we want per-notebook environment. @sourishkrout I could use your advice here - where in the code is the mechanism to manage notebook metadata? This metadata in particular should probably not be sent to the serializer since its irrelevant for serialization.
 - Environment is maybe not deleted on exit. I tried to make this happen as part of the `dispose` cycle, but it's still janky and needs further testing


src/extension/kernel.ts
```diff
@@ -237,14 +248,170 @@ export class Kernel implements Disposable {
...
+          if(code) {
+            // TODO: do something with this code
+            resolve(false)
+          }
+
+          resolve(true)
+        })
+
+        if(!interactive) {
+          const MIME_TYPES_WITH_CUSTOM_RENDERERS = ['text/plain']
```
degrammer, line 299: I would suggest to move this to constants file

mxsdev: Before merge this whole function will be brought out, probably into its own file. Constants will be put at top-level as part of that process 👍

src/extension/runner.ts
```diff
@@ -0,0 +1,378 @@
...
+   * there is a reference to it in this object and its already been disposed of
+   */
+  protected registerChild(d: DisposableAsync) {
+    this.children.push(new WeakRef(d))
+  }
+}
+
+// TODO: this should not be allowed to run anything if the parent (GrpcRunner)
+// runs `close` 
+export class GrpcRunnerProgramSession implements IRunnerProgramSession {
```
christian-bromann, line 158: I feel like a Runme session should be embedded as a VS Code terminal similar how we did it in https://github.com/stateful/vscode-runme/blob/main/src/extension/terminal/terminal.ts as a POC. When opening a terminal, it would create a connection and when closing it it would shut down. I think it makes it easier for us to send/receive commands from the extension API and prevents things like dangling sessions.

mxsdev: `IRunnerProgramSession` does implement `Pseudoterminal`, and should essentially act like a VSCode terminal - for example, upon running `open` it starts the program, and on `close` tries to close the program (need upstream changes to make this work better though).

Did you have something particular in mind about the implementation that's much different than what a VS Code terminal would do?

src/extension/runner.ts
```diff
@@ -0,0 +1,378 @@
...
+// TODO: handle connection failures better!!
+export class GrpcRunner implements IRunner {
+  protected readonly client: RunnerServiceClient
+  protected transport: GrpcTransport
+
+  private children: WeakRef<DisposableAsync>[] = []
+
+  constructor() {
+    this.transport = new GrpcTransport({
+      host: 'unix:///tmp/runme/runme.sock',
```
sourishkrout, line 81: We are already using `unix:///tmp/runme.sock` and `localhost:8080`. let's not introduce a third option.

sourishkrout:
> * `InputData` sent in the `sendRawInput` function (located in `runner.ts > GrpcRunnerProgramSession`) is completely empty by the time it gets to `runnerService.Execute` on the server. It's getting totally dropped, and this is a big blocker for interactive support. Do you have any idea why this might be? @sourishkrout please test and let me know if you have the same issue/any thoughts here, I have absolutely no idea what could be causing this...

I think i see the problem and have a hunch why but I'm not sure if I'm looking at the right "issue". What are the steps to reproduce for you @mxsdev?

src/extension/runner.ts
mxsdev: Oh yeah that's my bad, will fix 👍

mxsdev:
@sourishkrout 

I included a line that printed the `InputData` as-is in the `runme` go code, and it was always empty - didn't matter what I ran. Just tested `node ./scripts/stdin.js` and have the same issue.

I use `zsh` on my system, if that's relevant.

sourishkrout:
> @sourishkrout
> 
> I included a line that printed the `InputData` as-is in the `runme` go code, and it was always empty - didn't matter what I ran. Just tested `node ./scripts/stdin.js` and have the same issue. Also have the same issue with a `read` command in a bash script.
> 
> I use `zsh` on my system, if that's relevant.

I believe the problem you're running up against here is that the Pty is not fully implemented. In other words, individual keystrokes result in individual calls of input data (and likely individual requests) and aren't being buffered until, let's say RETURN is being pressed what would tell the Pty to send the buffered input stream off to the server.

That's the fun of implementing the Pseudoterminal, now one has to define and implement the different user interactions with a underlying terminal.

adambabik:
> I get the following trace when running Execute. I think what's happening here is that the program has exited, so the stream has ended, but srv.Recv() still gets called one last time, resulting in this "rpc Canceled" error. If this is not a client-side issue, I think we should make an issue about it on the runme repo.

Ideally, you'd like to notify the runner service that you want to drop the connection and let it do it. For example, when running a long running process like an HTTP server, try to kill it first by sending `0x03` byte (you likely need to start it with TTY (`ExecuteRequest.tty`). Then, the server should close the connection and send the exit code of the process. Alternatively, you can handle the `rpc Canceled` error if that's what you expect.

> Is there any way to request a program exit other than sending an interrupt? Currently if the user requests a command stop, the client sends an EOF, but then the program continues running. I think maybe that the program should try to stop if its still running during an EOF, since an interrupt can be cancelled by the program, but something like cmd.Stop() cannot - what do you think?

Can you send the command and how you execute it (with TTY or not), please? Typically, the command should exit on its own. If you want to interrupt it, currently, you must start it with TTY and send `0x03` and `0x04` bytes (at least for bash).

Regarding the second part, when a program handles signals and ignores them, the client can drop the connection and the program will exit. You should try that with https://github.com/stateful/runme/pull/127 as it might have been broken.

Another solution would be to add a field to `ExecuteRequest` which would indicate a force stop.

adambabik:
> Alternatively, you can handle the rpc Canceled error if that's what you expect.

There is actually another way that is preferable. I don't know exactly how the TS interface looks like but when you start a bi-directional stream, you can close the send direction with `stream.CloseSend()` in Go. This will notify the server that the client no longer wants to send anything. When this happens, the runner service will close the process' stdin, which usually terminates the whole process.

Check out [`service_test.go`](https://github.com/stateful/runme/blob/6c45d2824168ef8428d2f5784b8b85fa58a8eba3/internal/runner/service_test.go) to see different variants. Overall, the following should always hold:
- If you start an execution with TTY, try to interrupt the running command and close the send direction.
- If you start an execution without TTY, the only reliable way is to cancel, unless the program exits when the stdin is closed (`bash` works that way, for example), then closing the send direction is enough.
- If the executing command is unresponsive, cancelling will force it to stop.

sourishkrout:
> * I added the main functionality to `kernel.ts` as an inline async function. Much of this is copy/pasted and/or adapted from `taskExecutor` and `shellExecutor`. I want to bring this to its own file/method, but its not clear exactly how. I basically see two ways to proceed:
>   
>   1. Add a new `executor` function which runs the given runner. This is basically what I've done already, but results in lots of duplication/furthering the aforementioned issue of cementing the `executor` model.
>   2. Create an abstraction - my idea is something like an abstract class `ExecutionNode`, so for the runner it will be `RunnerExecutionNode`. For now it will be implemented rather simply (maybe it flatly produces some kind of "`ExecutionResult`" interface), but I think this is a good step towards things like multiple output stages, cleaner input utilization, chaining, etc.
> * Basically, if I understood correctly, in our 1:1 you were describing a desire for a kind of "middle ground" solution between a simplistic function (executor) and a full-blown solution. I think I can implement a class (`ExecutionNode`) which satisfies this desire and also can be built upon going forward. Maybe the exact abstraction described in (ii) isn't the one you want to go with, but in any case I don't think it will be a significant development effort to implement it. But I need your approval before going this way, as we could easily just do (i) and then work towards (ii) later/in follow up PRs, and I know this feature is time-sensitive.

Just a quick comment on this one: I'd like to keep us focused on getting this to a place where it's shippable. Meaning, not try to do too many things in one PR. So my advice would be to potentially keep it simple and see if we can design a test suite that will help us with a refactor later. If there's an incremental step forward we can make, feel free to go ahead. From what I understand, I believe your thoughts are on the right track. Again, I'd suggest to keep it simple for now and capture notes for another PR to evolve executors down the line.

mxsdev:
Should be good on my end, basically waiting for [#131](https://github.com/stateful/runme/issues/131) and [#130](https://github.com/stateful/runme/issues/130) before we can review/merge 👍

sourishkrout:
> PR should be basically ready, contingent on runme [#134](https://github.com/stateful/runme/pull/134) being merged.

cool 🤘. i'll start my review in a little bit. hopefully will finish before EoD.

mxsdev:
> cool 🤘. i'll start my review in a little bit. hopefully will finish before EoD.

FYI, if you want to do testing, you'll need to link the newest generated files from that PR, since there's been a change to the proto defs.

sourishkrout:
Overall this is a big step in the right direction. Let's figure out what the definition of "done" is for this PR.

sourishkrout (changes requested):

sourishkrout:
Let's look into `message: 'gRPC Server closed output stream unexpectedly!'` after interactive execution completes.

List of incomplete functionality outside of this PR:
- [ ] Environment Variables inherited from host-process
- [ ] Use correct $SHELL
- [ ] User-prompts for ENV variables
- [ ] Deep integrations: Vercel/deno
- [ ] Terminal dimensions (not high priority)

src/extension/executors/runner.ts
```diff
@@ -0,0 +1,163 @@
...
+  const { interactive, mimeType, background } = annotations
+  
+  const program = await runner.createProgramSession({
+    // TODO: make this work on windows/if bash isn't installed/etc
+    programName: execKey,
+    environment,
+    script: {
+      type: 'script', script
+    },
+    envs: [
```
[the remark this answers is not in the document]
mxsdev: I added this as part of the environment creation, so that this doesn't end up overwriting any changes to environment made by scripts. Let me know if you'd rather it be on every execution like you had it 👍

mxsdev:
@sourishkrout Anything more preventing this merge?

sourishkrout (approved):
> @sourishkrout Anything more preventing this merge?

Actually just tested this with the canonical `README.md` and the expected low-hanging fruit fixes work 👍 . Everything else I added to the new ticket: https://github.com/stateful/vscode-runme/issues/212.

Let's merge 💯!

PR #207, opened 2023-02-14
Fix annotations not saving if file has markdown blocks

mxsdev:
If there's a non-code block before a code block, it'll break from the checking loop pre-maturely. So I think we have to use `continue` here instead.


sourishkrout (approved):
👍 LGTM

sourishkrout:
This is what happens when I don't write flow control functional-style. The one time I use loops.

mxsdev:
And sebastian never made that mistake again!

sourishkrout:
> And sebastian never made that mistake again!

It's all Monads from here on out

PR #209, opened 2023-02-16
Launch runme server as part of the extension

degrammer:
Default settings:

Screenshot 2023-02-16 at 6 42 43 PM

Ensure you have a runme binary under the specified folder (by default .bin) at the root level


sourishkrout:
Taking a look at this right now. Looks like the code needs some adjustments.

sourishkrout:
@degrammer please take a look at https://github.com/stateful/vscode-runme/pull/211 which will make this PR work end-to-end. It does require the runme v0.6 pre-release.

Overall, the code appears to treat `runme server` like a CLI while it's not. Every "server" has a Command Line Interface which is basically just tracing its logs. I'd suggest to invert the all enclosing promise to "monitor" the logs for the successful listening output. For error handling, it's likely more important to check for non-zero error codes etc.

src/extension/cli/server.ts
```diff
@@ -0,0 +1,86 @@
...
+                // TODO: Parse data to detect if there is any error, for now, simulating nothing failed.
+                const isError = false
+                if (isError) {
+                    return reject(new ServerError(`Server failed, reason: ${data.toString()}`))
+                }
+
+                const msg = data.toString()
+                try {
+                  const log = JSON.parse(msg)
+                  if ('started listening' === log?.['msg']) {
```
mxsdev: I don't think it's a good idea to test for success by `'msg'` field, since that's subject to change (could theoretically be localized at some point for example) and has no guarantee of uniqueness. 

If you want to test for success, it's probably better to test for the `addr` field, which currently only gets sent on the "listening" message.

src/extension/cli/server.ts
```diff
@@ -0,0 +1,86 @@
...
+                    this.#retryOnFailure && isError ?
+                        this.#start() :
+                        resolve(log)
+                  }
+                } catch (err: any) {
+                  reject(new ServerError(msg))
```
mxsdev: As babik noted, non-runme messages can be logged in any format, and those may not necessarily be unrecoverable errors. I don't think you want to reject if there's a parse error, maybe just log it to console.

src/extension/cli/server.ts
```diff
@@ -0,0 +1,86 @@
...
+
+import ServerError from './serverError'
+
+export interface IServerConfig {
+    assignPortDynamically?: boolean
+    retryOnFailure?: boolean
+    maxNumberOfIntents: number
+}
+
+class Server implements Disposable {
```
mxsdev: I feel like the name `Server` is extremely vague. How would you feel about something like `RunmeServer`?

src/extension/cli/serverError.ts
```diff
@@ -0,0 +1,8 @@
+class ServerError extends Error {
+    constructor(message: string) {
+        super(message)
+        this.name = 'runme server error'
```
mxsdev:
```suggestion
        this.name = 'RunmeServerError'
```

JS error name convention

src/extension/cli/server.ts
```diff
@@ -0,0 +1,86 @@
...
+    #runningPort: number
+    #process: ChildProcessWithoutNullStreams | undefined
+    #binaryPath: string
+    #retryOnFailure: boolean
+    #intent: number
+    #maxNumberOfIntents: number
+
+    constructor(options: IServerConfig) {
+        const config = vscode.workspace.getConfiguration('runme.server')
+        this.#runningPort = config.get<number>('port')!
```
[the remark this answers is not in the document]
sourishkrout: If we don't have a proper port set, we should probably throw an error... the only way we should get to empty port is when the settings.json is invalid.

src/extension/cli/server.ts
sourishkrout: Totally open to that, however, just wanted to echo what I said before: lets not use the term `CLI` anywhere here since Runme's Command Line Interface is not involved whatsoever.

src/extension/cli/server.ts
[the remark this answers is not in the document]
mxsdev: Agreed. Basically I think we should be explicit about what to do when there's no port, rather than telling TS to just ignore it.

src/extension/cli/server.ts
```diff
@@ -0,0 +1,86 @@
...
+        this.#intent ++
+        if (this.#maxNumberOfIntents < this.#intent) {
+            return new ServerError(`Cannot start server, reached max number of intents: ${this.#maxNumberOfIntents}`)
+        }
+        const binaryExists = await fs.access(this.#binaryPath)
+            .then(() => true, () => false)
+        if (!binaryExists) {
+            throw new ServerError('Cannot find server binary file')
+        }
+        return new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
sourishkrout: To perhaps add a frame to think about this: I'd look at it more like a scheduler, e.g. if `runme server` ran inside of a container, `class Server` is basically it's scheduler (something like containerd or dockerd). A more basic and naive one.

src/extension/cli/server.ts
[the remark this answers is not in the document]
degrammer: I have some ideas for this, using Zod and more generic for our entire configuration system.

src/extension/cli/server.ts
degrammer: Fixed, changed to use addr

src/extension/cli/server.ts
degrammer: Changed to the specific use case when rejecting makes sense (when starting the server) other cases will be logged (under a Runme setting flag)

src/extension/cli/server.ts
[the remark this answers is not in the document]
degrammer: Re-organized as discussed today

src/extension/cli/server.ts
degrammer: Changed as suggested.

src/extension/server/runmeServer.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    constructor(options: IServerConfig) {
+        const config = workspace.getConfiguration('runme.server')
+        this.#runningPort = config.get<number>('port')!
+        this.#loggingEnabled = config.get<boolean>('enableLogger', false)
+        this.#binaryPath = path.join(__dirname, '../', config.get<string>('binaryPath')!)
+        this.#retryOnFailure = options.retryOnFailure || false
+        this.#maxNumberOfIntents = options.maxNumberOfIntents
+        this.#intent = 0
+        this.#address = `${SERVER_ADDRESS}:${this.#runningPort}`
```
sourishkrout, line 37: I'd make sure this is really an int here... cmd-line injections since it's set by the user.

src/extension/serializer.ts
```diff
@@ -291,8 +292,10 @@ export class GrpcSerializer extends SerializerBase {
   protected ready: ReadyPromise
 
   static async initGrpc(): Promise<ParserServiceClient> {
+    const config = workspace.getConfiguration('runme.server')
```
sourishkrout, line 295: I'd suggest to move all the config parsing logic into a helper/utility that returns a validated set of values which can be used wherever this config's required

sourishkrout (changes requested):
This is looking aweomse! Some small things.

src/extension/serializer.ts
degrammer: Moved to a configuration utility plus zod schema validation

src/extension/server/runmeServer.ts
degrammer: Moved to zod schema

sourishkrout (approved):
The validation changes look 👍 .

Two things I realized in priority order; I suggest merging anyways though:

1. If you have multiple windows open we will attempt to start multiple instances on the same socket. I guess, in the short-term, we can just give up starting the server if it fails x consecutive times? An alternative would be to hit a GRPC status endpoint on a running socket to verify it's indeed already running.
2. When I `sudo lsof -nP -i4TCP:8084` and `sudo kill <pid>` the respective process after it has come up. It won't restart.

I'd suggest to track both as separate tickets. Especially 1.) since it might require development on the server.

sourishkrout:
@mxsdev if you haven't already now's a good time to review... we'd like to merge soon.

tests/extension/extension.test.ts
```diff
@@ -11,6 +11,12 @@ vi.mock('../../src/extension/grpc/client', () => ({
 }))
 
 test('initializes all providers', async () => {
+  const configValues = {
+    binaryPath: '.bin/runme'
+  }
+  vi.mocked(workspace.getConfiguration).mockReturnValue({
+    get: vi.fn().mockImplementation((config: string) => configValues[config])
```
mxsdev, line 18:
```suggestion
    get: vi.fn((config: string) => configValues[config])
```

FYI you can pass the implementation as an argument to `vi.fn`!

mxsdev:
Good to merge imo, only thing is that one test case which could give a false positive.

sourishkrout:
> Good to merge imo, only thing is that one test case which could give a false positive 👍

Spun off https://github.com/stateful/vscode-runme/issues/224 and https://github.com/stateful/vscode-runme/issues/225 which we can work on later.

src/extension/server/runmeServer.ts
```diff
@@ -0,0 +1,97 @@
...
+                }
+            })
+        })
+    }
+
+    async launch(): Promise<object | number | RunmeServerError> {
+        try {
+            return await this.#start()
+        } catch (e) {
+            if (this.#retryOnFailure && this.#maxNumberOfIntents > this.#intent) {
```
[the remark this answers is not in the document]
degrammer: Retry on failure is the best effort here before initializing the extension, retrying on close event can happen after the extension was already initialized, not sure how we can implement a retry mechanism without affecting the user experience while using VSCode, I added a change to display an error message to the user in case the process is closed, see an example:

https://user-images.githubusercontent.com/4001529/220452718-854bfb3f-633b-4113-8399-be3f952e2c6c.mp4

tests/extension/server/runmeServer.test.ts
```diff
@@ -0,0 +1,41 @@
...
+    } as any)
+
+    test('Should try 2 times before failing', async () => {
+        const server = new Server({
+            retryOnFailure: true,
+            maxNumberOfIntents: 2
+        })
+        const serverLaunchSpy = vi.spyOn(server, 'launch')
+        try {
+            await server.launch()
```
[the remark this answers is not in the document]
degrammer: Fixed, I used rejects.toBeInstanceOf instead.

tests/extension/extension.test.ts
degrammer: Fixed

src/extension/server/runmeServer.ts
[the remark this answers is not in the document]
mxsdev: Yeah that's fine for now, I was just mentioning it in reference to solving #225.

mxsdev (approved):

PR #144, opened 2023-01-24
Provide insights for Runme improvements

sourishkrout:
This PR will turn on a basic-level of telemetry to allow the team to gain better insights into Runme's usage.

Any information emitted is pseudo-anonymized (no PII whatsoever) and the telemetry emitter will respect VS Code's global "send no telemetry" (id: `telemetry.telemetryLevel`) setting.

Please file a ticket or [find us on Discord](https://discord.gg/stateful) if you have any questions. Also, please review the paragraph about telemetry in the bottom section of [Runme's docs](https://runme.dev/docs/intro).

PR #143, opened 2023-01-24
Update Extension dependencies

christian-bromann:
This patch updates the following dependencies:

```
 @octokit/rest                        ^19.0.5  →    ^19.0.7
 @types/node                        ^18.11.15  →  ^18.11.18
 @types/vscode                        ^1.71.0  →    ^1.74.0
 @types/yargs                        ^17.0.19  →   ^17.0.20
 @typescript-eslint/eslint-plugin     ^5.48.0  →    ^5.49.0
 @typescript-eslint/parser            ^5.48.0  →    ^5.49.0
 @vercel/client                      ^12.2.24  →    ^12.3.2
 @vercel/frameworks                   ^1.1.15  →     ^1.2.4
 @vitest/coverage-c8                  ^0.26.3  →    ^0.28.1
 @vscode/vsce                         ^2.15.0  →    ^2.16.0
 @vscode/webview-ui-toolkit             1.1.0  →      1.2.1
 @wdio/cli                             ^8.1.2  →     ^8.2.3
 @wdio/local-runner                    ^8.1.2  →     ^8.2.3
 @wdio/mocha-framework                 ^8.1.2  →     ^8.2.3
 eslint                               ^8.29.0  →    ^8.32.0
 eslint-import-resolver-typescript     ^3.5.2  →     ^3.5.3
 eslint-plugin-import                 ^2.26.0  →    ^2.27.5
 expect-webdriverio                    ^4.0.1  →     ^4.1.0
 fork-ts-checker-webpack-plugin       ^7.2.14  →     ^7.3.0
 lit                                   ^2.5.0  →     ^2.6.1
 undici                               ^5.14.0  →    ^5.16.0
 vercel                              ^28.10.2  →   ^28.13.1
 vitest                               ^0.25.8  →    ^0.28.1
 webdriverio                           ^8.1.2  →     ^8.2.3
```

These are all patch or minor updates so I don't expect any issues with it. With #141 these updates will happen automatically.

PR #226, opened 2023-02-21
Runner service implementation feature parity

mxsdev:
- [x] Run `background=true` as background tasks
- [x] User-prompts (UI) for ENV variables
- [x] Unify cmds/script prep?
- [x] Allow identification of grpc-created terminals

Needs changes in runme service:

- [x] Curl-ing img (last cell in `examples/README.md`) with notebook-native png renderer crash-hangs [#143](https://github.com/stateful/runme/issues/143)
- [x] Cat-ing LICENSE (in `examples/README.md`) does not work ~~[#144](https://github.com/stateful/runme/issues/144)~~ [#155](https://github.com/stateful/runme/issues/155)

### DIFF
--- a/src/extension/executors/runner.ts
+++ b/src/extension/executors/runner.ts
@@ -52,6 +52,10 @@ export async function executeRunner(
 
     const userValue = await promptUserForVariable(key, value, hasStringValue)
 
+    if(userValue === undefined) {
+      return false
+    }
+
     envs[key] = userValue
 
     /**
@@ -72,7 +76,7 @@ export async function executeRunner(
     script: {
       type: 'commands', commands
     },
-    envs: Object.entries(envs).map(([k, v]) => `${k}="${v}"`),
+    envs: Object.entries(envs).map(([k, v]) => `${k}=${v}`),
     cwd,
     tty: interactive
   })
--- a/src/extension/executors/runner.ts
+++ b/src/extension/executors/runner.ts
@@ -10,7 +10,8 @@ import {
   TaskPanelKind,
   tasks,
   NotebookCellExecution,
-  TextDocument
+  TextDocument,
+  ExtensionContext
 } from 'vscode'
 
 import { OutputType } from '../../constants'
@@ -27,6 +28,7 @@ const BACKGROUND_TASK_HIDE_TIMEOUT = 2000
 const MIME_TYPES_WITH_CUSTOM_RENDERERS = ['text/plain']
 
 export async function executeRunner(
+  context: ExtensionContext,
   runner: IRunner,
   exec: NotebookCellExecution,
   runningCell: TextDocument,
@@ -111,6 +113,10 @@ export async function executeRunner(
       program.close()
     })
 
+    context.subscriptions.push({
+      dispose: () => program.close()
+    })
+
     await program.run()
   } else {
     const taskExecution = new Task(
@@ -131,6 +137,10 @@ export async function executeRunner(
     }
 
     const execution = await tasks.executeTask(taskExecution)
+
+    context.subscriptions.push({
+      dispose: () => execution.terminate()
+    })
 
     exec.token.onCancellationRequested(() => {
       try {
@@ -190,7 +200,7 @@ export async function executeRunner(
       setTimeout(
         () => {
           closeTerminalByEnvID(RUNME_ID)
-          return resolve(true)
+          resolve(true)
         },
         BACKGROUND_TASK_HIDE_TIMEOUT
       )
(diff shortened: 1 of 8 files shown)


src/extension/executors/runner.ts
```diff
@@ -10,22 +10,25 @@ import {
...
 import { closeTerminalByEnvID } from './task'
-import { getShellPath } from './utils'
+import { getCommandExportExtractMatches, getShellPath, promptUserForVariable } from './utils'
 
 const LABEL_LIMIT = 15
+const BACKGROUND_TASK_HIDE_TIMEOUT = 2000
 const MIME_TYPES_WITH_CUSTOM_RENDERERS = ['text/plain']
 
 export async function executeRunner(
+  context: ExtensionContext,
```
sourishkrout, line 31: Side note: one of the reasons why I think this the function-abstraction of Executors won't suffice long-term. Not part of this PR, though.

src/extension/executors/runner.ts
```diff
@@ -46,11 +88,9 @@ export async function executeRunner(
     programName: getShellPath(execKey),
     environment,
     script: {
```
sourishkrout, line 90: Consider renaming `script` to `exec` / `executable` or something like that. Script appears contradictory to commands being the name of one of many types.

src/extension/executors/utils.ts
```diff
@@ -46,21 +98,18 @@ export async function retrieveShellCommand (exec: NotebookCellExecution) {
...
-    if (placeHolder.startsWith('$(') && placeHolder.endsWith(')')) {
+
+  for (const { hasStringValue, key, match, type, value } of exportMatches) {
+    if (type === 'exec') {
       /**
        * evaluate expression
        */
-      const expression = placeHolder.slice(2, -1)
-      const expressionProcess = cp.spawn(expression, {
+      const expressionProcess = cp.spawn(value, {
```
sourishkrout, line 112: Shouldn't this use the Runner APIs versus spawning child processes? Consider addressing in follow up issue/PR.

src/extension/utils.ts
```diff
@@ -106,8 +111,8 @@ export function getKey(runningCell: vscode.TextDocument): keyof typeof executor
  * treat cells like like a series of individual commands
  * which need to be executed in sequence
  */
-export function getCmdShellSeq(cellText: string, os: string): string {
-  const trimmed = cellText
+export function getCmdSeq(cellText: string): string[] {
```
sourishkrout, line 114: Once the TUI is using the Runner to execute commands we should consider GRPC exposing APIs to prep an input as a sequence of commands so both Ext/TUI plausible work the same way. Let's make a ticket for this if we haven't already.

sourishkrout (changes requested):
Left some comments in the code. From a functional point of view, it appears that "export-only" cells have some unindented behavior.

Please see this clip: https://i.imgur.com/jqYi1yH.mp4

src/extension/executors/utils.ts
mxsdev: This feature uses `getCommandExportExtractMatches` directly - it does not call `retrieveShellCommand`.

sourishkrout: great 👌 , thanks for clarifying!

mxsdev:
@sourishkrout

Tested with the main example readme and every script worked for me. 

Could you test against really quick with the latest changes? Thanks!

src/extension/utils.ts
mxsdev: https://github.com/stateful/runme/issues/157

sourishkrout:
> @sourishkrout
> 
> Tested with the main example readme and every script worked for me.
> 
> Could you test against really quick with the latest changes? Thanks!

Confirmed. Great job 👏 

Another thing I noticed is that `closeTerminalOnSuccess` does not seem to have an effect? Just on my end, @mxsdev?

sourishkrout (approved):
👍 LGTM

mxsdev:
I was able to fix tests, but for the following line in `tests/extension/executors/utils.test.ts`:

```ts
vi.mock('../../../src/extension/utils', () => ({
  replaceOutput: vi.fn(),
}))
```

I was not able to use `importActual` as it would result in the process hanging. I guess this is due to cyclic dependency, but that's [supposed to be fixed](https://github.com/vitest-dev/vitest/issues/2567), so idk.

In the meantime I just mocked the method, but this is less than ideal since there's no need to mock it. Lmk if you think this is worth an issue @sourishkrout.

Anyways, this is still good to merge! 🎉

PR #230, opened 2023-02-23
Break out bringing up GRPC server from consuming a service

sourishkrout:
For flexible development and testing until this is fully finalized closer to release.


sourishkrout:
cc @degrammer & @mxsdev: not a significant change just needed the ability to control things individually for targeted testing.

Issue #233, opened 2023-02-24
Output duplication on consecutive runs due to underlying bug in vscode


sourishkrout:
This is related to https://github.com/microsoft/vscode/issues/173577 and, for the time being, can be fixed by calling `clearOuput` ahead of replaceOutput. Essentially replicating the behavior.

![image](https://user-images.githubusercontent.com/250527/221205238-d240d7ce-31c1-4332-bbb1-164a74c4af20.png)

Let's be sure to fix all instances of this issue.

PR #234, opened 2023-02-24
Until underlying bug's fixed

sourishkrout:
Fixes https://github.com/stateful/vscode-runme/issues/233.


degrammer (approved):
Tested and GLTM

sourishkrout:
> Tested and GLTM

Cool. The e2e test has surfaced an edge case. Looking into that before merging.

PR #248, opened 2023-02-28
Add new extension distribution with binaries

degrammer:
Introducing a new GitHub action for distributing the extension with the new Runme binaries.

Operating System Support Matrix:

- Linux ARM64, X64
- MacOS (Darwin ARM64, X64)
- Windows (win32-X64, win32-arm64)

It supports specific runme versions or use latest as default, non stable and draft releases are **not** supported

Screenshot 2023-02-28 at 12 00 38 PM

### DIFF
--- a/.github/scripts/downloadAndExtract.js
+++ b/.github/scripts/downloadAndExtract.js
@@ -5,11 +5,11 @@ import path from 'node:path'
 import fs from 'node:fs'
 import { parseArgs, promisify } from 'node:util'
 
-
 import tar from 'tar-fs'
 import unzipper from 'unzipper'
 
 const streamPipeline = promisify(pipeline)
+const VALID_FORMATS = ['zip', 'tar']
 const SCHEMA = {
     downloadUrl: {
         type: 'string',
@@ -61,7 +61,7 @@ const cleanCliDownload = async (cliDownloadFolder) => {
 }
 
 const downloadBinary = async (downloadUrl, type, binaryDestination, binaryName) => {
-    if (!['zip', 'tar'].includes(type)) {
+    if (!VALID_FORMATS.includes(type)) {
         throw new Error(`Invalid type ${type}, it must be zip or tar file`)
     }
 
@@ -72,7 +72,7 @@ const downloadBinary = async (downloadUrl, type, binaryDestination, binaryName) 
     if (!binaryName) {
         throw new Error('Missing binary name')
     }
-    
+
     const cliFilePath = path.resolve(binaryDestination)
     await cleanCliDownload(cliFilePath)
     await fsp.mkdir(cliFilePath, { recursive: true })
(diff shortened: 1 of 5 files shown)


.github/scripts/downloadAndExtract.js
```diff
@@ -0,0 +1,125 @@
...
+
+    console.info("🗑️ Cleaning CLI Download folder...")
+
+    await fsp.rm(cliDownloadFolder, { recursive: true, force: true })
+
+    console.info("✅ CLI Download folder cleared!")
+}
+
+const downloadBinary = async (downloadUrl, type, binaryDestination, binaryName) => {
+    if (!['zip', 'tar'].includes(type)) {
```
christian-bromann, line 64: Maybe we can make `['zip', 'tar']` a const as well?

.github/workflows/binaries.yml
```diff
@@ -0,0 +1,460 @@
+name: Release extension with new runme binaries
```
christian-bromann, line 1: if this is a release workflow maybe call the file `release.yml`?

.github/workflows/binaries.yml
```diff
@@ -0,0 +1,460 @@
...
+          tag: ${{ env.GIT_TAG }}
+          prerelease: ${{ github.event.inputs.releaseChannel == 'edge' }}
+      
+      - name: 🐛 Debug Build
+        uses: stateful/vscode-server-action@v1
+        if: failure()
+        with:
+          timeout: "300000"
+
+
```
christian-bromann, line 460:
```suggestion
```

.github/workflows/binaries.yml
degrammer: yes, the idea is to eventually call this file release.yml once we are 100% sure everything works and remove the current release.yml file.

degrammer:
> Some comments, more minor. I think scripts are simple enough that we are fine going without having tests for them but this would be always a good thing to have. Up to you.

@christian-bromann  thank you, yeah, I will track the missing tests under https://github.com/stateful/vscode-runme/issues/250

package.json
```diff
@@ -386,7 +386,9 @@
     "test:lint:fix": "eslint src tests --ext ts --fix",
     "test:unit": "vitest -c ./vitest.conf.ts",
     "test:e2e": "wdio run ./tests/e2e/wdio.conf.ts",
-    "watch": "npm run build:dev -- --watch"
+    "watch": "npm run build:dev -- --watch",
+    "vscode:prepublish": "npm run prepare-binary",
```
mxsdev, line 390: We might want to run this script on `postinstall` as well as we move from wasm. Maybe not in this PR, but something to keep note of.

Note also that since there is no host process identification in the current scripts, we'll need a third script later on to identify the host's platform and then download the proper binary.

mxsdev (approved):
Didn't do extensive testing but the code looks fine (a lot of it is adapted from what I wrote, so I'm familiar with it). `downloadAndExtract.js` also worked for me.

sourishkrout (approved):
Once we have the parser released, let's be sure to add a dev-workflow to this as per @mxsdev's suggestion. Otherwise 👍 .

PR #254, opened 2023-03-01
Remove Publishing Edge releases to Open VSX

degrammer:


degrammer:
Fixes https://github.com/stateful/vscode-runme/pull/254

PR #257, opened 2023-03-01
Improve path configuration/resolution

mxsdev:
Implements the feedback I gave in #245 .

In summary:

 - Binary path configuration now points to the binary itself, rather than the enclosing folder
 - There is one singular function for finding the binary path, `getBinaryPath`
 - The extension root Uri is passed down as a Uri, rather than as a string (avoids unnecessary (re)-parsing)
 - Relative binary paths now act relative to the current workspace folder, rather than relative to extension folder

### DIFF
--- a/__mocks__/vscode.ts
+++ b/__mocks__/vscode.ts
@@ -1,3 +1,4 @@
+import path from 'node:path'
 import { readFileSync } from 'node:fs'
 import { join } from 'node:path'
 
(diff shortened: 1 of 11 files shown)


sourishkrout:
> Seems to work as intended on OSX, @sourishkrout if you could test on Windows - I don't have access to a windows PC at the moment.

Will do. Will report back before EoD.

sourishkrout (changes requested):
Not quite working yet: The zod schema further up on configuration.ts resets the default to `bin` which threw me for a loop.

Otherwise looks a look cleaner 👍 .

sourishkrout (approved):
👍 LGTM

PR #259, opened 2023-03-02
One session per window

mxsdev:
Runs one session per window, with incremental port assigning.

Note that there will be a race condition here (between deciding the port and running the server, the port can become occupied). The only way to avoid this would be putting the port assignment code on go, or by adding resilience to binding errors in the server process launch logic.

### DIFF
--- a/src/extension/server/runmeServer.ts
+++ b/src/extension/server/runmeServer.ts
@@ -35,16 +35,16 @@ class RunmeServer implements Disposable {
     #acceptsInterval: number
     events: EventEmitter
 
-    constructor(extBasePath: string, options: IServerConfig) {
-        this.#port = getPortNumber()
-        this.#loggingEnabled = enableServerLogs()
-        this.#binaryPath = getBinaryPath(extBasePath, process.platform)
-        this.#retryOnFailure = options.retryOnFailure || false
-        this.#maxNumberOfIntents = options.maxNumberOfIntents
-        this.#intent = 0
-        this.#acceptsIntents = options.acceptsConnection?.intents || 50
-        this.#acceptsInterval = options.acceptsConnection?.interval || 200
-        this.events = new EventEmitter()
+    constructor(extBasePath: Uri, options: IServerConfig) {
+      this.#port = getPortNumber()
+      this.#loggingEnabled = enableServerLogs()
+      this.#binaryPath = getBinaryPath(extBasePath, process.platform)
+      this.#retryOnFailure = options.retryOnFailure || false
+      this.#maxNumberOfIntents = options.maxNumberOfIntents
+      this.#intent = 0
+      this.#acceptsIntents = options.acceptsConnection?.intents || 50
+      this.#acceptsInterval = options.acceptsConnection?.interval || 200
+      this.events = new EventEmitter()
     }
 
     dispose() {
--- a/src/extension/server/runmeServer.ts
+++ b/src/extension/server/runmeServer.ts
@@ -171,6 +171,10 @@ class RunmeServer implements Disposable {
         await this.acceptsConnection()
         return addr
     }
+
+    private _port() {
+      return this.#port
+    }
 }
 
 export default RunmeServer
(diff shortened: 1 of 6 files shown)


src/extension/server/runmeServer.ts
```diff
@@ -90,10 +87,14 @@ class RunmeServer implements Disposable {
             throw new RunmeServerError('Cannot find server binary file')
         }
 
+        while (!(await isPortAvailable(this.#port))) {
+          this.#port++
```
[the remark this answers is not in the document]
mxsdev: @christian-bromann It is a while loop so it will keep incrementing until it finds a port not in use.

I wouldn't have any problems with using a random port, but we already decided to go with an incremental approach.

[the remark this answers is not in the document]
christian-bromann: gotcha

mxsdev:
This is working by the way, just currently wrangling with tests

tests/extension/server/runmeServer.test.ts
```diff
@@ -38,6 +45,28 @@ suite('Runme server spawn process', () => {
...
+          retryOnFailure: true,
+          maxNumberOfIntents: 2,
+        }
+      )
+
+      vi.mocked(fs.access).mockResolvedValueOnce()
+      // @ts-expect-error
+      vi.mocked(fs.stat).mockResolvedValueOnce({
+        isFile: vi.fn().mockReturnValue(true)
+      })
```
christian-bromann, line 62: Not sure what the TS error is but I usually just cast things to any
```suggestion
      vi.mocked(fs.stat).mockResolvedValueOnce({
        isFile: vi.fn().mockReturnValue(true)
      } as any)
```

mxsdev: TS error is because `fs.stat` expects a much larger struct. But yeah, casting as any here is better!

src/extension/utils.ts
```diff
@@ -310,3 +311,7 @@ export function replaceOutput(
 export function getGrpcHost() {
   return `${SERVER_ADDRESS}:${getPortNumber()}`
 }
+
+export async function isPortAvailable(port: number): Promise<boolean> {
+  return (await getPort({ port })) === port
```
degrammer, line 316: I'm wondering if we should restrict this a little bit, by specifying a range of ports that are desirable to use, starting with **7863**, also to reduce the chances that after the first iteration it will find a port being used.

```
import getPort, {portNumbers} from 'get-port';

await getPort({port: portNumbers(7863, 8000)})
```

degrammer (approved):
LGTM, left small comment

sourishkrout:
> Note that there will be a race condition here (between deciding the port and running the server, the port can become occupied). The only way to avoid this would be putting the port assignment code on go, or by adding resilience to binding errors in the server process launch logic.

I think the risk is acceptably low since "the race" is between the same user's interactions. If multiple windows open all at once, the retry loop should make this work eventually.

sourishkrout (approved):
👍 LGTM. Tested on Windows with both WSL and non-WSL.

src/extension/utils.ts
sourishkrout: Having a ceiling on the port number is not mutually exclusive with the incremental approach. In a small follow-up PR we could add that. However, I'd suggest providing for more headroom than 8000.

mxsdev: @sourishkrout Ah I see. The code as provided though will assign randomly in the range. 

If the port exceeds the ceiling, would you basically just error at that point?

sourishkrout: Oh, sorry... didn't see at first that the API implemented that. Predictable ports will make examples of interop between extension and CLI easier to demo until we figure out "service discovery." That's the main motivation why I'm attached to the incremental assignment approach.... for now.

sourishkrout: > If the port exceeds the ceiling, is the idea that you would error at that point?

Yep.

PR #263, opened 2023-03-03
Make run with Runme easier by not requiring a full git setup

### LINKED ISSUE #258: UriHandler: Instead of normalization, accept both git and https git repo references
The UriHandler should prefer https over git (if present) refs for public repos. That way it won't matter whether or not a user has git configured and/or if `github.com` is added to their list of known hosts.

christian-bromann:
fixes #258

### DIFF
--- a/src/extension/handler/uri.ts
+++ b/src/extension/handler/uri.ts
@@ -7,7 +7,7 @@ import { TelemetryReporter } from 'vscode-telemetry'
 
 import {
     getProjectDir, getTargetDirName, waitForProjectCheckout, getSuggestedProjectName, writeBootstrapFile,
-    parseParams
+    parseParams, gitSSHUrlToHTTP
 } from './utils'
 
 let NEXT_TERM_ID = 0
@@ -15,7 +15,7 @@ const REGEX_WEB_RESOURCE = /^https?:\/\//
 
 export class RunmeUriHandler implements UriHandler {
     async handleUri (uri: Uri) {
-        console.log(`[Runme] triggered RunmeUriHandler with ${uri}`)
+        console.log(`[Runme] triggered RunmeUriHandler with ${uri}!!!`)
         const params = new URLSearchParams(uri.query)
         const command = params.get('command')
 
@@ -103,10 +103,10 @@ export class RunmeUriHandler implements UriHandler {
         fileToOpen: string
     ) {
         progress.report({ increment: 0, message: 'Cloning repository...' })
-        const terminal = window.createTerminal(`Runme Terminal #${NEXT_TERM_ID++}`, '/bin/sh')
+        const terminal = window.createTerminal(`Runme Terminal #${NEXT_TERM_ID++}`)
         terminal.show(true)
 
-        terminal.sendText(`git clone ${repository} ${targetDirUri.fsPath}`)
+        terminal.sendText(`git clone ${gitSSHUrlToHTTP(repository)} ${targetDirUri.fsPath}`)
         const success = await new Promise<boolean>(
             (resolve) => waitForProjectCheckout(fileToOpen, targetDirUri.fsPath, repository, resolve))
 
--- a/src/extension/handler/uri.ts
+++ b/src/extension/handler/uri.ts
@@ -7,13 +7,11 @@ import { TelemetryReporter } from 'vscode-telemetry'
 
 import {
     getProjectDir, getTargetDirName, waitForProjectCheckout, getSuggestedProjectName, writeBootstrapFile,
-    parseParams, gitSSHUrlToHTTPS
+    parseParams
 } from './utils'
 
 let NEXT_TERM_ID = 0
 const REGEX_WEB_RESOURCE = /^https?:\/\//
-
-const sleep = (ms = 200) => new Promise((resolve) => setTimeout(resolve, ms))
 
 export class RunmeUriHandler implements UriHandler {
     async handleUri (uri: Uri) {
@@ -108,21 +106,9 @@ export class RunmeUriHandler implements UriHandler {
         const terminal = window.createTerminal(`Runme Terminal #${NEXT_TERM_ID++}`)
         terminal.show(true)
 
-        terminal.sendText(`git clone ${gitSSHUrlToHTTPS(repository)} ${targetDirUri.fsPath}`)
+        terminal.sendText(`git clone ${repository} ${targetDirUri.fsPath}`)
         const success = await new Promise<boolean>(
             (resolve) => waitForProjectCheckout(fileToOpen, targetDirUri.fsPath, repository, resolve))
-
-        /**
-         * We always checkout the project using HTTPS to avoid issues with git not being set up.
-         * However in case the user provided a ssh url we want to switch back to the original
-         * origin url so in case git was set up the user can work with the right git connection.
-         */
-        if (repository.startsWith('git@')) {
-            terminal.sendText(`cd ${targetDirUri.fsPath}`)
-            await sleep()
-            terminal.sendText(`git remote set-url origin ${repository}`)
-            await sleep()
-        }
 
         if (!success) {
             return terminal.dispose()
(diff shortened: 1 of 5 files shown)


src/extension/handler/utils.ts
```diff
@@ -157,6 +157,15 @@ export function getSuggestedProjectName (repository: string) {
     return
 }
 
+export function gitSSHUrlToHTTP(repositoryUrl: string) {
```
sourishkrout, line 160: This should probably be `HTTPS`

sourishkrout (changes requested):
We should accept git-ssh repo refs and https-based git repos refs separately. If I add a "badge" to my internal repo, I want to use git-ssh and it's safe to assume folks who have access to the private repo have ssh-auth configured.

In other words, let's validate but not do normalization whatsoever.

christian-bromann:
@sourishkrout are you suggesting that we should try with a git-ssh url if provided and if that fails we should fallback to `https`?

christian-bromann:
> Up to the "author" to decide what best fits the use-case

That is how it is currently implemented. I figured to apply this change anyway because of the fact what happens if git is not set up. So how about the following idea: always pull source using HTTPS but if a git url is provided change the origin afterwards back to the original using another terminal command.

What do you think?

sourishkrout:
I just tested it:

```
URI vscode://stateful.runme?command=setup&repository=git@github.com:stateful/blog-examples.git&fileToOpen=node-streams/README.md
Actual Remote: https://github.com/stateful/blog-examples.git
Expected Remote: git@github.com:stateful/blog-examples.git
```

```
URI: vscode://stateful.runme?command=setup&repository=https://github.com/stateful/blog-examples.git&fileToOpen=node-streams/README.md
Actual Remote: Exception Thrown
Expected Remote: https://github.com/stateful/blog-examples.git
```

So right now, it appears broken when I test it end-to-end. The repository URL should only get validated (to prevent injections). That's what I would expect.

We don't need to check if it worked. It's on the author to pick their auth strategy.

christian-bromann:
@sourishkrout yes, my bad, wasn't testing that case. However everything seems to work as expected now. Please have another test.

src/extension/handler/uri.ts
```diff
@@ -7,15 +7,17 @@ import { TelemetryReporter } from 'vscode-telemetry'
...
 
 let NEXT_TERM_ID = 0
 const REGEX_WEB_RESOURCE = /^https?:\/\//
 
+const sleep = (ms = 200) => new Promise((resolve) => setTimeout(resolve, ms))
+
 export class RunmeUriHandler implements UriHandler {
     async handleUri (uri: Uri) {
-        console.log(`[Runme] triggered RunmeUriHandler with ${uri}`)
+        console.log(`[Runme] triggered RunmeUriHandler with ${uri}!!!`)
```
sourishkrout, line 20: `!!!`?

src/extension/handler/uri.ts
```diff
@@ -103,13 +105,25 @@ export class RunmeUriHandler implements UriHandler {
...
-        const terminal = window.createTerminal(`Runme Terminal #${NEXT_TERM_ID++}`, '/bin/sh')
+        const terminal = window.createTerminal(`Runme Terminal #${NEXT_TERM_ID++}`)
         terminal.show(true)
 
-        terminal.sendText(`git clone ${repository} ${targetDirUri.fsPath}`)
+        terminal.sendText(`git clone ${gitSSHUrlToHTTPS(repository)} ${targetDirUri.fsPath}`)
         const success = await new Promise<boolean>(
             (resolve) => waitForProjectCheckout(fileToOpen, targetDirUri.fsPath, repository, resolve))
 
+        /**
```
sourishkrout, line 115: I need clarification on why this is necessary... One can't check out a private git repo without authorization, no matter whether it's HTTPS or SSH. The only upshot HTTPS has over SSH is for public repos where GIT authentication (forgoes authorization) requires pubkeys to be set up is performed as a result. HTTPS only does authentication if authorization is required which is just a fundamental protocol difference that plays nicely with public repos.

src/extension/handler/uri.ts
```diff
@@ -103,13 +105,25 @@ export class RunmeUriHandler implements UriHandler {
         fileToOpen: string
     ) {
         progress.report({ increment: 0, message: 'Cloning repository...' })
-        const terminal = window.createTerminal(`Runme Terminal #${NEXT_TERM_ID++}`, '/bin/sh')
+        const terminal = window.createTerminal(`Runme Terminal #${NEXT_TERM_ID++}`)
         terminal.show(true)
 
-        terminal.sendText(`git clone ${repository} ${targetDirUri.fsPath}`)
+        terminal.sendText(`git clone ${gitSSHUrlToHTTPS(repository)} ${targetDirUri.fsPath}`)
```
sourishkrout, line 111: Not 100% sure, however, I'm not seeing the HTTPS URL being checked for validity to prevent injections here.

sourishkrout (changes requested):
It completes the test cases now as expected. However, I commented on issue with the implementation. Please see review comments.

> ```
> URI vscode://stateful.runme?command=setup&repository=git@github.com:stateful/blog-examples.git&fileToOpen=node-streams/README.md
> Actual Remote: https://github.com/stateful/blog-examples.git
> Expected Remote: git@github.com:stateful/blog-examples.git
> ```
> 
> 
>     
>   
> 
> ```
> URI: vscode://stateful.runme?command=setup&repository=https://github.com/stateful/blog-examples.git&fileToOpen=node-streams/README.md
> Actual Remote: Exception Thrown
> Expected Remote: https://github.com/stateful/blog-examples.git
> ```

src/extension/handler/uri.ts
christian-bromann: > The only upshot HTTPS has over SSH is for public repos where GIT authentication (forgoes authorization) requires pubkeys to be set up is performed as a result.

Exactly that. With this patch cloning the repo will also work for folks that don't have git set up. But as you said, might be problematic for folks that try to clone a private repo. I guess the question is, who will use this more:

- folks that don't have git setup
- folks that try to checkout a private repository

I hunch was the first but if you say you like to focus on latter more I can disregard the PR.

Another option is to allow setting another url query parameter to tag the git resource as private in which case we don't do this transition.

src/extension/handler/uri.ts
christian-bromann: There are two places:
- input parsing to prevent injection: [`parseParams`](https://github.com/stateful/vscode-runme/blob/ac53278c1041d6afb29c2c4d95648a50ca295fab/src/extension/handler/utils.ts#L177-L198) (which is called right at the beginning of the uri handler
- validating format of passed in url: [`getSuggestedProjectName`](https://github.com/stateful/vscode-runme/blob/ac53278c1041d6afb29c2c4d95648a50ca295fab/src/extension/handler/utils.ts#L138-L158)

src/extension/handler/uri.ts
sourishkrout: > Exactly that. With this patch cloning the repo will also work for folks that don't have git set up. But as you said, might be problematic for folks that try to clone a private repo. I guess the question is, who will use this more:
> 
> * folks that don't have git setup
> * folks that try to checkout a private repository
> 
> I hunch was the first but if you say you like to focus on latter more I can disregard the PR.
> 
> Another option is to allow setting another url query parameter to tag the git resource as private in which case we don't do this transition.

As I said before, we want to support both but not "programagically". It's not the extension's place to figure out what git repo reference (git vs https) to use. That's up to the author of the link/badge to decide. So in essence, if GIT is provided in the URI, use GIT, if HTTPS is provided, use HTTPS (as long as they are valid respectively).

We should then add a paragraph to the feature's docs to explain what to look out for.

christian-bromann: Gotcha, I miss-understood then from the [original issue](https://github.com/stateful/vscode-runme/issues/258#issue-1606894447) which mainly asked to improve the UX for users without git setup. I can have VS Code print a helpful error message for these users then.

sourishkrout (approved):
The test cases work as expected now. Let's merge 👍 .

PR #266, opened 2023-03-03
Provide better feedback on Form validation

degrammer:
Small fix to provide better form validation while the user types

better-validation

### DIFF
--- a/src/client/components/annotations.ts
+++ b/src/client/components/annotations.ts
@@ -186,6 +186,7 @@ export class Annotations extends LitElement {
       type="text"
       value="${text}"
       @change="${this.#onChange}"
+      @keyup="${ this.#onChange}"
       placeholder=${placeHolder}
       size="50"
       class="annotation-item"


christian-bromann:
One suggestion: if no value is entered can it autofill it with its default value?

sourishkrout (approved):
👍  LGTM

degrammer:
> One suggestion: if no value is entered can it autofill it with its default value?

@christian-bromann we could, but I'm wondering if that's a good idea, since the default value will be applied anyway? if we add that, it will end up in the annotation being added all the time to the markdown file.

sourishkrout:
> > One suggestion: if no value is entered can it autofill it with its default value?
> 
> @christian-bromann we could, but I'm wondering if that's a good idea, since the default value will be applied anyway? if we add that, it will end up in the annotation being added all the time to the markdown file.

let's skip it this time. defaults are local to the extension. something we need to fix but don't have the time for right now.

PR #269, opened 2023-03-04
Better handle kernel selection error

sourishkrout:
Prompt users to select a notebook kernel instead of the raw error message.

image


christian-bromann (approved):
👍

Issue #225, opened 2023-02-21
Not restarting on termination


sourishkrout:
When I `sudo lsof -nP -i4TCP:8084` and `sudo kill <pid>` the respective process after it has come up. It won't restart.

PR #281, opened 2023-03-06
Auto-Restart server on `onClose`

mxsdev:
Makes runme server resistant against closing, as long as it is caught by the `onClose` callback.

Might want to think about polling/regular health checks in case something more dubious occurs. (Long-term)

Should be functional, just need to write/fix tests at this point.

### DIFF
--- a/src/extension/extension.ts
+++ b/src/extension/extension.ts
@@ -31,12 +31,13 @@ export class RunmeExtension {
     const kernel = new Kernel(context)
     const grpcSerializer = kernel.hasExperimentEnabled('grpcSerializer')
     const grpcServer = kernel.hasExperimentEnabled('grpcServer')
+    const grpcRunner = kernel.hasExperimentEnabled('grpcRunner')
     const server = new RunmeServer(context.extensionUri, {
       retryOnFailure: true,
       maxNumberOfIntents: 2,
-    }, !grpcServer)
+    }, !grpcServer, grpcRunner)
+    grpcRunner && kernel.useRunner(new GrpcRunner(server))
     const serializer = grpcSerializer ? new GrpcSerializer(context, server) : new WasmSerializer(context)
-    kernel.hasExperimentEnabled('grpcRunner') && kernel.useRunner(new GrpcRunner(server))
 
     /**
      * Start the Runme server
--- a/src/extension/extension.ts
+++ b/src/extension/extension.ts
@@ -34,19 +34,15 @@ export class RunmeExtension {
     const server = new RunmeServer(context.extensionUri, {
       retryOnFailure: true,
       maxNumberOfIntents: 2,
-    })
+    }, !grpcServer)
+    const serializer = grpcSerializer ? new GrpcSerializer(context, server) : new WasmSerializer(context)
+    kernel.hasExperimentEnabled('grpcRunner') && kernel.useRunner(new GrpcRunner(server))
+
     /**
      * Start the Runme server
      */
-
     try {
-      if (grpcServer) {
-        await server.launch()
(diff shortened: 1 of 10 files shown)


sourishkrout:
K, I'm going to test this end-to-end shortly.

sourishkrout:
Worked on Mac and Windows (both WSL and regular).

sourishkrout (approved):
Impl looks good to me 👍 . In a future rewrite we should throw the events out and use observables instead. :-D

src/extension/extension.ts
```diff
@@ -31,22 +31,19 @@ export class RunmeExtension {
...
     /**
      * Start the Runme server
      */
-
     try {
-      if (grpcServer) {
-        await server.launch()
-        server.events.on('closed', () => {
-          window.showErrorMessage(`Runme server is not longer running,
-          please reload the window to start the server again`)
```
degrammer, line 47: There is any chance we should keep this message in case the retry fails?

degrammer (approved):
LGTM, minor comment

src/extension/extension.ts
mxsdev: Ok, so upon further investigation, if `launch` fails, then an error will be thrown, and you'll get the error message below, saying basically the same thing.

If we want to log a message when the server has to restart anyway, we can explore that.

degrammer: ahh if that's the case 👍🏽

sourishkrout:
This PR fixes https://github.com/stateful/vscode-runme/issues/225.

src/extension/kernel.ts
```diff
@@ -316,15 +318,23 @@ export class Kernel implements Disposable {
...
+
+        runner.createEnvironment(
+          // copy env from process naively for now
+          // later we might want a more sophisticated approach/to bring this serverside
+          Object.entries(process.env).map(([k, v]) => `${k}=${v || ''}`)
+        ).then(env => {
+          if(this.runner !== runner) { return }
+          this.environment = env
+        })
+      })
```
christian-bromann, line 337: We can use `async/await` here to remove indentation:

```suggestion
      this.runnerReadyListener = runner.onReady(async () => {
        this.environment = undefined

        const env = await runner.createEnvironment(
          // copy env from process naively for now
          // later we might want a more sophisticated approach/to bring this serverside
          Object.entries(process.env).map(([k, v]) => `${k}=${v || ''}`)
        )

        if(this.runner !== runner) { return }
        this.environment = env
      })
```

src/extension/runner.ts
```diff
@@ -98,23 +106,46 @@ export interface IRunnerProgramSession extends DisposableAsync, Pseudoterminal {
...
+
+  private initRunnerClient(transport?: GrpcTransport) {
+    this.deinitRunnerClient()
+    this.client = new RunnerServiceClient(transport ?? this.server.transport())
+    this._onReady.fire()
+  }
+
+  protected static assertClient(client: RunnerServiceClient|undefined): asserts client {
+    if(!client) {
+      throw new Error('Client is not active!')
```
christian-bromann, line 140: Could this be something a user is exposed to? In that case showing an `window.showErrorMessage` might be more appropiate.

sourishkrout: Fair point. I assumed it's caught "top of stack," but that might not be the case always, @mxsdev. However, since it's all in the Runner (which is default deactivated), let's do this in a follow-up issue. Let's get this merged. We need to get a release out today.

sourishkrout:
@mxsdev please "port" the unresolved comment to a new issue. Thanks!

PR #286, opened 2023-03-08
Improve gRPC runner error handling

mxsdev:
Significantly improves the error handling situation for gRPC runner. Included is:

 - Handle case where server is disconnected prematurely and close terminal (previously terminal task would hang)
 - If grpc runner fails, show error message to user
 - If grpc environment creation fails, show error message to user

### DIFF
--- a/src/extension/kernel.ts
+++ b/src/extension/kernel.ts
@@ -267,7 +267,7 @@ export class Kernel implements Disposable {
       this.runner &&
       !(hasPsuedoTerminalExperimentEnabled && terminal)
     ) {
-      const runScript = (execKey: 'sh'|'bash' = 'bash') => executeRunner(
+      const runScript = async (execKey: 'sh'|'bash' = 'bash') => await executeRunner(
         this.context,
         this.runner!,
         exec,
@@ -277,7 +277,8 @@ export class Kernel implements Disposable {
         environmentManager
       )
         .catch((e) => {
-          console.error(e)
+          window.showErrorMessage(`Internal failure executing runner: ${e.message}`)
+          console.error('Internal failure executing runner', e.message)
           return false
         })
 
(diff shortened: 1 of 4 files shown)


src/extension/kernel.ts
```diff
@@ -323,17 +324,22 @@ export class Kernel implements Disposable {
...
+            // later we might want a more sophisticated approach/to bring this serverside
+            Object.entries(process.env).map(([k, v]) => `${k}=${v || ''}`)
+          )
+
           if(this.runner !== runner) { return }
           this.environment = env
-        })
+        } catch(e: any) {
+          window.showErrorMessage(`[Runme] Failed to create environment for gRPC Runner: ${e.message}`)
+          console.error('Failed to create gRPC Runner environment', e)
```
christian-bromann, line 341: I've seen `[Runme] ` prefixes mostly in our internal logs, no strong preference though, feel free to disregard.

```suggestion
          window.showErrorMessage(`Failed to create environment for gRPC Runner: ${e.message}`)
          console.error('[Runme] Failed to create gRPC Runner environment', e)
```

src/extension/runner.ts
```diff
@@ -440,6 +451,13 @@ export class GrpcRunnerProgramSession implements IRunnerProgramSession {
   }
 
   open(): void {
+    // Workaround to force terminal to close if opened after early exit
+    // TODO(mxs): find a better solution here
```
christian-bromann, line 455: Should we port this into an issue before we land the PR?

src/extension/kernel.ts
```diff
@@ -277,7 +277,8 @@ export class Kernel implements Disposable {
         environmentManager
       )
         .catch((e) => {
-          console.error(e)
+          window.showErrorMessage(`Internal failure executing runner: ${e.message}`)
+          console.error('Internal failure executing runner', e.message)
```
christian-bromann, line 281:
```suggestion
          console.error('[Runme] Internal failure executing runner', e.message)
```

src/extension/kernel.ts
christian-bromann: Screenshot 2023-03-14 at 14 06 55

christian-bromann (commented):
Some minor comments, otherwise LGTM 👍

src/extension/runner.ts
mxsdev: I'll add an issue for it.

I don't think it's very high priority, basically just the `_onDidClose` gets fired twice for an interactive task. I think most things that listen to that are idempotent so it should be fine.

mxsdev: #286

PR #291, opened 2023-03-10
Use gRPC health check for runme server

mxsdev:
Detect whether gRPC server exists using `grpc/grpc` healthcheck rather than using the parser client directly. This is more in line with gRPC standard practice and gives us more options down the line.

### DIFF
--- a/src/extension/server/runmeServer.ts
+++ b/src/extension/server/runmeServer.ts
@@ -5,11 +5,11 @@ import path from 'node:path'
 import { ChannelCredentials } from '@grpc/grpc-js'
 import { GrpcTransport } from '@protobuf-ts/grpc-transport'
 import { Disposable, Uri, EventEmitter } from 'vscode'
+
 import {
   HealthCheckRequest,
   HealthCheckResponse_ServingStatus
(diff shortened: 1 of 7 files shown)


src/extension/server/runmeServer.ts
```diff
@@ -5,12 +5,15 @@ import path from 'node:path'
 import { ChannelCredentials } from '@grpc/grpc-js'
 import { GrpcTransport } from '@protobuf-ts/grpc-transport'
 import { Disposable, Uri, EventEmitter } from 'vscode'
+import {
+  HealthCheckRequest,
+  HealthCheckResponse_ServingStatus
+} from '@buf/grpc_grpc.community_timostamm-protobuf-ts/grpc/health/v1/health_pb'
```
sourishkrout, line 11: More of a nit, don't love bleeding the full module name outside the "grpc" namespace. I'd suggest to import and export them just like we have with other generated types.

sourishkrout (changes requested):
See namespacing comment. Granted it passed functional testing on both Windows and Mac, good to go 👍 .

src/extension/server/runmeServer.ts
mxsdev: See 4432482

sourishkrout (approved):
👍 LGTM

Issue #289, opened 2023-03-10
Duplicate cell execution error


degrammer:
When running multiple cell executions, we display this unfriendly error message to the user:

<img width="533" alt="Screenshot 2023-03-10 at 10 26 20 AM" src="https://user-images.githubusercontent.com/4001529/224360124-00296f23-d97b-4695-a55b-2c09033fa00b.png">

- Figure out if there is a better way to handle this situation, it could be by disabling the configuration buttons if the cell is in an execution state or by exploring with [supportExecutionOrder](https://github.com/stateful/vscode-runme/blob/d2dfda7beb5259843cebf7cdedc20ffdd6226a08/src/extension/kernel.ts#L63)

Steps to reproduce the issue:
- Run a background task
- Click configure while running
- You should get the above error

mxsdev:
Did some investigation. The only way I found in the VSCode API to modify a cell's outputs is through a running execution, and you are not allowed more than 1 running execution at a time. `supportsExecutionOrder` appears to be strictly aesthetic (showing the number in the bottom left corner).

So the only way I saw to fix this was to keep track of the running execution ourselves, and force all modifications of the output to be done in a unified way. 

I was able to implement this in #304

PR #304, opened 2023-03-14
Allow concurrent outputs

mxsdev:
Solves #289 

Allows multiple parts of the code to concurrently manage a cell's outputs without having to worry about executions:

<p align="center">

</p>

To do this, I've unified output handling to a new file `cell.ts`, which acts as a single source of truth for all output handling.

Changes are synchronized using a new `Mutex` utility class, which essentially acts as a dynamic promise queue that ensures mutations to outputs are done in sequence. I am open to alternative patterns for accomplishing this.

### The API

The important parts are mostly in `cell.ts`, which has a `NotebookCellOutputManager` class. It is expected that one is created per each cell, where this would be the new central location for dealing with that cell's outputs. In particular, it has the responsibility of generating outputs for each output type, based on the metadata of that cell.

The expectation is that all matters of outputs are dealt with through this class. For example, to toggle the annotations output, you would do `outputs.toggleOutput(OutputType.annotations)`.

As more complex output behavior is needed, this class should provide the internal tools to make this happen by extending the API. All of the race conditions should be dealt with.

### Persisting Output Data

There is a broader move here from sending the data to an output through the json, to persisting it with the cell's metadata. This is necessary since cells will potentially be re-rendered as the user e.g. opens the annotation window, toggles outputs, etc.

The solution given is very far from excellent, and I am all ears to suggestions. But hopefully, it should work for now. 

There is a data duplication/caching issue (same data is sent over the messaging channel as is persisted through metadata cache), and it is perhaps confusing how you set the metadata and then have to manually call a refresh. 

In the future, we should introduce a broader abstraction to handle/unify state synchronization and persistence in outputs. Bonus points if we can make generating `NotebookCellOutput` objects part of this as well. Interesting if we can learn from marquee/tangle here, since it's a very similar problem.

## Known issues:

 - [x] A user manually removing outputs is not handled. Ideally we can listen to an event for this, if not we have to figure it out lazily.
 - [x] Non-gRPC execution code is untouched
 - [x] gRPC execution code needs to be unified for vercel & deno outputs
 - [x] Deno output should hide embedded terminal
 - [x] No unit tests (yet)
 
Other known issues:

 - Deno output won't persist promoted (if you toggle annotations, promotion status gets reset)


degrammer (approved):
This looks amazing, LGTM, tested, it feels super smooth

src/extension/cell.ts
```diff
@@ -0,0 +1,284 @@
...
+} from 'vscode'
+
+import { OutputType } from '../constants'
+import { CellOutputPayload, Serializer } from '../types'
+import { Mutex } from '../utils/sync'
+
+import { getAnnotations, replaceOutput, validateAnnotations } from './utils'
+
+export class NotebookCellManager {
+  private data = new WeakMap<NotebookCell, NotebookCellOutputManager>()
```
christian-bromann, line 20: I would advocate to use actual private fields, e.g.
```suggestion
  #data = new WeakMap<NotebookCell, NotebookCellOutputManager>()
```

unless you need the object for unit testing purposes.

src/extension/kernel.ts
```diff
@@ -258,13 +265,19 @@ export class Kernel implements Disposable {
     })
   }
 
-  public async createCellExecution(cell: NotebookCell): Promise<NotebookCellExecution> {
-    return this.#controller.createNotebookCellExecution(cell)
+  public async createCellExecution(cell: NotebookCell): Promise<RunmeNotebookCellExecution> {
+    return await this.cellManager.createNotebookCellExecution(cell)
+    // return this.#controller.createNotebookCellExecution(cell)
```
christian-bromann, line 270: still needed?
```suggestion
```

src/extension/provider/annotations.ts
```diff
@@ -49,41 +45,7 @@ export class AnnotationsProvider implements NotebookCellStatusBarItemProvider {
...
-      ])
-    } catch (e: any) {
-      if (e.message.toString().includes('controller is NOT associated')) {
-        return this.handleNotebookKernelSelection()
-      }
-      window.showErrorMessage(e.message)
-    } finally {
-      exec?.end(true)
-    }
+    await (await this.kernel.getCellOutputs(cell)).toggleOutput(OutputType.annotations)
```
christian-bromann, line 48: maybe easier to read:
```suggestion
    const outputCell = await this.kernel.getCellOutputs(cell)
    await outputCell.toggleOutput(OutputType.annotations)
```

src/utils/sync.ts
```diff
@@ -0,0 +1,44 @@
+/* Mutex provides mutex feature by building a promise chain.
+  const m = new Mutex();
+  await m.withLock(async () => {
+    // critical section
+  })
+
+  https://github.com/golang/vscode-go/blob/fa820d45d7b9217e949c21ffd13268e20c7e0ee4/src/utils/mutex.ts#L8
+*/
+export class Mutex {
+	private mutex = Promise.resolve()
```
christian-bromann, line 10: Can we format the file, seems like the indentation is sometimes 8 and sometimes 2

christian-bromann (commented):
Some comments, but overall LGTM 👍

src/extension/kernel.ts
mxsdev: Nope, thanks for catching!

sourishkrout:
Something for next week - not super high priority short-term: Tried to review this branch but notebooks won't load for me for now. Assuming the final touches leading up to v1 led to inconsistencies and merge conflicts.

mxsdev:
@sourishkrout 

I rebased against main, this should be working now. (Problem was that TLS wasn't implemented on this branch)

mxsdev:
Now integrated with the notebook terminal 👍 

Also changed the "Open Terminal" button to just be "Terminal," and it now acts as a toggle across all output modes. I think this is better UX, let me know your thoughts!

sourishkrout:
Let's perhaps get all the remaining Notebook Terminal stuff merged & shipped before releasing this one into `edge` for a break-in period. Makes sense, @mxsdev?

mxsdev:
@sourishkrout 

Sounds good to me!

PR #308, opened 2023-03-16
Add new Terminal Component

degrammer:
This PR adds a new Terminal Component.
It needs to be enabled under the **runme.terminal** extension settings.


mxsdev:
I'll take a look at this once I get a chance. Did you add anything else since we last discussed?

degrammer:
> I'll take a look at this once I get a chance. Did you add anything else since we last discussed? @degrammer

@mxsdev  nothing new from what we last discussed that's worth to mention, added support to unicode11 characters and the serializer addon that will be used to keep the terminal state (as a separate PR)

examples/README.md
mxsdev, line 1: You should revert these changes or bring to separate PR

src/utils/configuration.ts
```diff
@@ -32,23 +36,35 @@ const getServerConfigurationValue = <T>(configName: keyof typeof configurationSc
     return defaultValue
 }
 
+const getRunmeTerminalConfiguration = <T>(configName: keyof typeof configurationSchema.terminal, defaultValue: T) => {
+    const configurationSection = workspace.getConfiguration(TERMINAL_SECTION_NAME)
+    const configurationValue = configurationSection.get<T>(configName)!
+    // Since a default value is configured, its safe to use parse here.
+
+    const parseResult = configurationSchema.terminal[configName].safeParse(configurationValue)
```
mxsdev, line 44:
```suggestion

    // Since a default value is configured, its safe to use parse here.
    const parseResult = configurationSchema.terminal[configName].safeParse(configurationValue)
```

Formatting 🤓

mxsdev (commented):
Yeah as you've mentioned, most of this is stuff either I wrote or have seen, so good by me, other than the things mentioned.

You'll need to fix tests before merging of course, if you have any issues with that let me know.

My main priority tomorrow is probably #304, so hopefully we can get progress on that by then.

examples/README.md
degrammer: Yeah, reverted

mxsdev (approved):
Good stuff, definitely want a follow-up PR adding more tests before we flip the switch 😉

Issue #312, opened 2023-03-17
Multiple shells / Shell detection


mxsdev:
Currently, we try to determine what shell to execute in based on the `SHELL` environment variable. This causes a few issues:

 - We cannot have different scripts execute in different shells, e.g. based on the language of the fenced code block
 - `SHELL` is not necessarily set on Windows, so this is a significant hurdle for windows support

VSCode internally has a list of `TerminalProvider`s which can be provided by extensions. We should:

 - Tap into this if possible
 - If not, develop a solution to detect shells on the host's system, ideally leveraging existing solutions/libraries

With this information, we can:

 - Give user the option to select a shell profile to run a command block in
 - Try to make a best guess for which shell to use by default (using language, user's preferred shell, etc)
 - Use gRPC runner on Windows

For windows support, runme will need to implement support for powershell and DOS. This should not be that difficult, see stateful/runme#173

PR #313, opened 2023-03-17
Hard disable grpc runner on windows

mxsdev:
Windows does not have a `$SHELL` environment variable, so it becomes our responsibility to detect the location of the shell. 

Basically, until #312 is solved, it's probably best to hard disable the gRPC runner on windows, since we don't know where the shell is.


degrammer (approved):
Lgtm

PR #324, opened 2023-03-21
fix `getLastStableRelease` for Windows

christian-bromann:
`os.platform()` returns in Windows `win32` while our download directory lists Windows artifacts as e.g. `0.6.5/runme_windows_x86_64.zip`.


degrammer (approved):
LGTM

Issue #290, opened 2023-03-10
Decommission ExperimentalTerminal


sourishkrout:
- Remove Experimental Terminal implementation
- Consider preserving Task definition in a way that makes sense with the GRPC Runner

PR #325, opened 2023-03-21
Decomission Experimental Terminal/Adapt Tasks for use with gRPC Runner

mxsdev:
Solves #290

Currently I didn't try to support tasks in non-grpc runner environments. ~~Tasks also don't use the same environment as the kernel (they don't have any environment persistence currently).~~

A lot of the stuff for tasks is copied from the runner executor. I'd like to unify the two but it's not super trivial. I see this is as a good POC for now, though I'm happy to take the time to solidify it if desired.


src/extension/provider/runmeTask.ts
```diff
@@ -63,36 +86,51 @@ export class RunmeTaskProvider implements TaskProvider {
...
-    const definition: RunmeTaskDefinition = {
-      type: 'runme',
-      filePath,
-      command,
-      closeTerminalOnSuccess,
-      isBackground,
-      cwd
+    const isBackground = options.isBackground || background
+
+    const runOpts: RunProgramOptions = {
```
sourishkrout, line 106: Do we have to "seed" the `envs` here with the host processes? Or does the runner take of that?

sourishkrout (changes requested):
Possible clarification.

src/extension/provider/runmeTask.ts
mxsdev: Oh yeah, we have to seed it, good call.

In the future I'd like to integrate this with the existing session.

sourishkrout: Makes sense.

mxsdev:
Added support for environments @sourishkrout

src/extension/provider/runmeTask.ts
mxsdev: Added support for environments and env population in 452eb96 @sourishkrout

sourishkrout (approved):
> A lot of the stuff for tasks is copied from the runner executor. I'd like to unify the two but it's not super trivial. I see this is as a good POC for now, though I'm happy to take the time to solidify it if desired.

👍 Good call. Let's circle back when the time's right.

PR #329, opened 2023-03-22
Remove git diff step

degrammer:


mxsdev (approved):
Glad we were able to figure that out.

Teamwork! 🙏

PR #331, opened 2023-03-23
Fix background tasks with gRPC runner

mxsdev:
Removed unneeded dispose call which was terminating the background program's execution prematurely.

My bad!


sourishkrout (approved):
👍

degrammer (approved):
LGTM

Issue #288, opened 2023-03-10
Update CLI provider to point to the shipped Runme CLI


christian-bromann:
Currently we have a CLI button for every code block that is suppose to use the Runme CLI to execute the command inside:

![cli](https://user-images.githubusercontent.com/731337/224189757-28cf1108-4e08-4daf-8454-58a7b14c9246.gif)

The button is using the `CliProvider` class which checks whether the CLI is installed to run the action. Now that we have the CLI shipped as part of the extension we can just point directly to it.

@sourishkrout happy to take a stab at this if you think this make sense.

sourishkrout:
This is about using the packaged binary's CLI versus looking inside the PATH (if installed via homebrew or other).

If so, please go ahead; I assume it's a subtle change.

Fyi, we will have to revisit the CLI button once the GRPC Runner integration wraps up. However, that's out of scope until we cross that bridge.

PR #335, opened 2023-03-24
Integrate CLI button with gRPC runner

mxsdev:
This fixes the CLI button for use with the gRPC runner. Needs version of binary including stateful/runme#202 (built from main)

Not sure whether to pass everything as command line parameters, or environment variables. I used environment variables because it looks cleaner (and allows users to easily run other commands in the same session), but a lot is then "hidden" from the user. Thoughts @sourishkrout ?

### DIFF
--- a/src/extension/commands/index.ts
+++ b/src/extension/commands/index.ts
@@ -1,13 +1,17 @@
 import path from 'node:path'
+import os from 'node:os'
 
 import {
   NotebookCell, Uri, window, env, NotebookDocument, TextDocument, ViewColumn,
   workspace, NotebookData, commands, NotebookCellData, NotebookCellKind
 } from 'vscode'
 
+import { getBinaryPath, getTLSDir, getTLSEnabled } from '../../utils/configuration'
 import { Kernel } from '../kernel'
 import { CliProvider } from '../provider/cli'
 import { getAnnotations, getTerminalByCell } from '../utils'
+import RunmeServer from '../server/runmeServer'
+import { GrpcRunnerEnvironment } from '../runner'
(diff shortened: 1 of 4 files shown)


src/extension/commands/index.ts
```diff
@@ -35,23 +39,58 @@ export function stopBackgroundTask (cell: NotebookCell) {
...
+      } else {
+        args.push('--insecure')
+      }
+
+      const runnerEnv = kernel.getRunnerEnvironment()
+      if(runnerEnv && runnerEnv instanceof GrpcRunnerEnvironment) {
+        envs['RUNME_SESSION'] = runnerEnv.getSessionId()
+      }
+    } else {
+      if (!await CliProvider.isCliInstalled()) {
```
mxsdev, line 74: Do you think we still need this @sourishkrout ?

I kept it around for compatibility reasons. But I should be good now to axe it in favor of the bundled binary, no?

sourishkrout: We should remove it. There's an issue for it: https://github.com/stateful/vscode-runme/issues/288. Let's do it in a separate PR.

sourishkrout:
> Not sure whether to pass everything as command line parameters, or environment variables. I used environment variables because it looks cleaner (and allows users to easily run other commands in the same session), but a lot is then "hidden" from the user. Thoughts @sourishkrout ?

I was inclined to do the same. Mainly because with further explanation the flags aren't particularly self-evident and in this integrated way it makes sense for it to "just work".

sourishkrout (approved):
Works great 👍

PR #336, opened 2023-03-25
Add notebook terminal flags

degrammer:


package.json
```diff
@@ -287,11 +287,23 @@
             "default": true,
             "markdownDescription": "Use TLS over GRPC server interface (strongly recommended - only disable if you know what you are doing!!!)"
           },
-          "runme.terminal.enabled": {
+          "runme.terminal.backgroundTask": {
+            "type": "boolean",
+            "scope": "machine",
+            "default": true,
+            "markdownDescription": "Enable Runme integrated terminal in Background tasks (Experimental)"
```
sourishkrout, line 294: let's replace all mention of "integrated" with "notebook". VS Code talks about it's native terminal as "integrated terminal". Just so we can avoid confusion.

sourishkrout: Lowercase `experimental` just to be consistent with others

sourishkrout (changes requested):
Two things:
1. Can we make the right terminal pop open when I click on a terminal associated with a background task?
2. Wondering if we can remove the notebook terminal associated with a background task when one clicks the "Stop Task" button. I believe we already have an event firing (https://github.com/stateful/vscode-runme/blob/bencho/notebook-terminal-flags/src/extension/provider/background.ts#L50) which we might be able to leverage.

sourishkrout:
Great stuff @degrammer! See my comments please. This is definitely enough for me to take screenshots over the wknd.

package.json
degrammer: Fixed

degrammer:
1. Done, please review and test since I did organized a little better the NotebookTerminal logic.
keep-state
2. Still missing, I will review it (it could be as a separate PR)

src/client/index.ts
```diff
@@ -59,6 +59,25 @@ export const activate: ActivationFunction = (context: RendererContext<void>) =>
...
+            if (content) {
+              context.setState(content)
+            }
+          })
+
+          const decodedNotebookTerminalContent = decodeURIComponent(
+            escape(window.atob(payload.output.content || ''))
+          )
+
+          terminalElement.setAttribute('initialContent',
```
mxsdev, line 77: Why does this PR contain a bunch of stuff to persist notebook content? It should really use the Notebook Serializer if possible.

Shouldn't we separate maintaining state into a separate PR?

src/utils/configuration.ts
```diff
@@ -47,31 +49,33 @@ const getServerConfigurationValue = <T>(configName: keyof typeof configurationSc
     return defaultValue
 }
 
-const getRunmeTerminalConfiguration = <T>(configName: keyof typeof configurationSchema.terminal, defaultValue: T) => {
+const getRunmeTerminalConfiguration = <T>(
```
mxsdev, line 52:
```suggestion
const getRunmeTerminalConfigurationValue = <T>(
```

Nit, consistency with other methods

src/extension/executors/runner.ts
```diff
@@ -146,8 +144,21 @@ export async function executeRunner(
...
+      const enableNotebookTerminal = mime === 'text/plain' && isNotebookTerminalFeatureEnabled('nonInteractive')
+      if (!enableNotebookTerminal) {
+        replaceOutput(exec, [new NotebookCellOutput([item])])
+      } else {
+        const json = NotebookTerminal.create(cellUUID, {
+          content: Buffer.concat(output).toString('base64'),
+          removeState: true
+        })
+
+        await replaceOutput(exec, [
```
mxsdev, line 156: By executing this as part of `handleOutput`, it will get called every time the output updates. 

I don't believe that is necessary. If you look above this code, you will see that the events for stdout get sent regardless of interactive status. 

Instead, I think you should early return from `handleOutput` if we're using the notebook terminal, and open the notebook terminal in this `if` branch, similar to the interactive case.

Even more ideally, you could open the terminal above the `if` branch.

src/extension/executors/notebookTerminal.ts
```diff
@@ -0,0 +1,24 @@
+import { NotebookCell, NotebookCellOutput, NotebookCellOutputItem, window } from 'vscode'
+
+import { Kernel } from '../kernel'
+import { replaceOutput } from '../utils'
+import { OutputType } from '../../constants'
+import NotebookTerminal from '../notebookTerminal'
+
+export async function NotebookTerminalExecutor(cell: NotebookCell, kernel: Kernel) {
```
mxsdev, line 8: This doesn't really fit in line with the rest of the "executors."

I think it would make more sense to have this as a method on the `NotebookTerminal` class - something like `openNotebookTerminal`. I feel like that would be more transparent as to what this code actually does.

mxsdev (changes requested):
I like the `NotebookTerminal` class - I think that's a good start for ensuring things like state/font are handled nicely! 👍 

I am a little confused by the overall implementation for the PR though. It looks like you added terminal state persistence so that you could re-render the terminal, but there should be no need to do that, since the stdout is sent over the messaging channel.

My thought was that for this PR, all you would need to do is only show the terminal (once) depending on the config flags, and that would all be in `executors/runner.ts`. Am I wrong in my assumption there? 

If not I feel like it would be a lot better to have terminal persistence in a separate PR, because this implementation seems incompatible with the XTerm Serializer plugin, and would probably need to be rewritten to support it.

degrammer:
Terminal flags moved to https://github.com/stateful/vscode-runme/pull/350/files

degrammer:
Closing this since it was done at https://github.com/stateful/vscode-runme/pull/351 and https://github.com/stateful/vscode-runme/pull/350

PR #337, opened 2023-03-25
Remove runner incompatible default

sourishkrout:
We will have fewer edge cases "without default" than introducing them starting their lifecycle in the extension.

There is more work to be done here but it's important to get this fix in for now.


mxsdev (approved):
This is fine, though I'm not sure what are the ramifications of having an empty name.

For example, the task list from #290 could have an empty listing.

Regardless I don't think empty names as a result of this PR would almost ever happen so I'm not too worried.

PR #348, opened 2023-03-27
Remove check for runme to be installed

### LINKED ISSUE #288: Update CLI provider to point to the shipped Runme CLI

mxsdev:
Fixes #288 

Removes parts of the code that require runme installed - we should just use the bundled/configured one instead.


degrammer (approved):
LGTM

PR #351, opened 2023-03-27
Notebook terminal persistence

mxsdev:
Persists notebook terminal output, so the contents of the terminal are retained, for example, through opening the config slide-out.

Also handles the `openTerminal` command @degrammer 

Still need to write/fix tests


src/extension/commands/index.ts
```diff
@@ -6,22 +6,44 @@ import {
...
+        try {
+          exec = await kernel.createCellExecution(cell)
+          exec.start(Date.now())
+
+          await replaceOutput(exec, terminalOutput)
+        } catch (e: any) {
+          window.showErrorMessage(e.message)
+        } finally {
+          exec?.end(true)
+          return
```
degrammer, line 40: This logic could be moved to a method exposed from here https://github.com/stateful/vscode-runme/blob/main/src/extension/notebookTerminal.ts

mxsdev: Yeah, I removed that class in this PR.

I'm happy to add it back at some point, but I felt like it was silly having an abstraction which just wraps one static method.

sourishkrout (approved):
👍 LGTM

PR #355, opened 2023-03-28
Use Task API for cloning project instead of timing out in Uri Handler

mxsdev:


christian-bromann:
@mxsdev let me know if you need any support getting this over the line.

mxsdev:
@christian-bromann 

Sure, I think I just need to fix tests. Not a super high priority which is why it's been sitting in draft so long.

src/extension/handler/uri.ts
```diff
@@ -111,23 +113,53 @@ export class RunmeUriHandler implements UriHandler {
...
+          }
+
+          /**
+           * only close terminal if execution passed and desired by user
+           */
+          if (e.exitCode === 0) {
+            resolve(true)
+          } else {
+            resolve(false)
+          }
```
christian-bromann, line 145:
```suggestion
          return resolve(e.exitCode === 0)
```

christian-bromann (approved):
Works great 🎉  tested it locally and on Windows 👍

PR #358, opened 2023-03-29
Fix relaunch behavior in runme server logic + window-specific certificates

mxsdev:
This PR does the following:

 - Fix relaunch behavior on close, which could sometimes be called in tandem with the relaunch behavior on error, causing two concurrent relaunches
 - Create TLS certificates in a different directory per-window rather than globally 
   - This behavior is overridden when a specific TLS directory is set - so the issue will still exist when the user specificies a directory, because runme CLI will always generate a new certificate on launch rather than reading an existing one)
 - Clean up some logic in the runme server to be iterative rather than recursive, which helps for debugging & code clarity


sourishkrout (approved):
👍 LGTM

PR #373, opened 2023-04-03
Fix notebook terminal resizing

mxsdev:
I don't think it's necessary to cache position to ensure not running fit, since it needs to be run every resize anyway.

This PR will be needed for implementing terminal (PTY) resizing.

### DIFF
--- a/src/client/components/terminal.ts
+++ b/src/client/components/terminal.ts
@@ -10,12 +10,6 @@ import { ClientMessages } from '../../constants'
 import { getContext } from '../utils'
 import { onClientMessage, postClientMessage } from '../../utils/messaging'
 
-
-interface IWindowSize {
-  width: number
-  height: number
-}
-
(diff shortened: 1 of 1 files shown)


sourishkrout:
Good point! I forgot to add this to the roadmap item

sourishkrout (changes requested):
Are integrated and notebook terminals competing here? Something seems off.

https://user-images.githubusercontent.com/250527/229798887-1e834f1a-8e4e-4596-8d0c-93f939c15d07.mp4

mxsdev:
@sourishkrout 

This PR is not for PTY resizing. Rather, it is a prerequisite for PTY resizing. Sorry if my language was ambiguous.

This PR is about resizing VSCode's window. It allows for word wrapping in this case, and for the notebook terminal itself to change size as a result. PTY resizing is for a follow-up PR.

sourishkrout (approved):
Sounds like you have a plan @mxsdev.

PR #374, opened 2023-04-03
Add copy button to notebook terminal

mxsdev:
Notebook terminal copy button

Adds copy button to notebook terminal


degrammer (approved):
LGTM,  we should eventually move this button to a separate component, in case we want to reuse it for other output items in the future.

PR #375, opened 2023-04-03
Force focus notebook terminal on window focus

mxsdev:
This forces a terminal focus when the window focues.

Before this, short clicks can sometimes fail to focus the terminal.

Unfortunately, this creates a "flicker effect" where you try to focus the terminal, it unfocuses, and then refocuses as a result of this PR. I figure this behavior is better than staying unfocused after clicking.

I am still not sure what is causing the issue in the first place, so this is effectively a work-around.


degrammer (approved):
LGTM, I agree this workaround is better than having the terminal unfocused

PR #376, opened 2023-04-04
Windows shell survey

sourishkrout:
Respectfully ask Windows users to share what default shell they are using so we can prioritize better.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
       {
         "command": "runme.surveyWinDefaultShell",
         "category": "Runme",
-        "title": "Prompt to partipciate in Windows Default Shell survey"
+        "title": "Prompt to participate in Windows Default Shell survey"
       },
       {
         "command": "runme.openAsRunmeNotebook",
--- a/package.json
+++ b/package.json
@@ -83,6 +83,11 @@
         "title": "Reset Environment Variables for current VS Code session"
       },
       {
+        "command": "runme.surveyWinDefaultShell",
+        "category": "Runme",
+        "title": "Prompt to partipciate in Windows Default Shell survey"
+      },
+      {
         "command": "runme.openAsRunmeNotebook",
         "category": "Runme",
         "title": "Open markdown as Runme",
--- a/src/extension/survey.ts
+++ b/src/extension/survey.ts
@@ -1,5 +1,5 @@
 import path from 'node:path'
-import { readFileSync, unlinkSync } from 'node:fs'
+import { mkdirSync, readFileSync, unlinkSync } from 'node:fs'
 
 import {
   Disposable,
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -82,6 +82,11 @@
         "category": "Runme",
         "title": "Reset Environment Variables for current VS Code session"
       },
+      {
+        "command": "runme.surveyWinDefaultShell",
+        "category": "Runme",
+        "title": "Prompt to partipciate in Windows Default Shell survey"
```
degrammer, line 88: Typo here: partipciate -> participate

degrammer, line 88: Typo here: partipciate -> participate

sourishkrout: Whooopsie. Just fixed. Thanks!

src/extension/survey.ts
```diff
@@ -0,0 +1,150 @@
...
+      'Don\'t ask again',
+      'Dismiss'
+    )
+    if (option === 'Dismiss') {
+      return
+    } else if (option !== 'OK') {
+      await this.#done()
+      return
+    }
+
```
mxsdev, line 79:
```suggestion
mkdirSync(this.#tempDir.fsPath, { recursive: true })
```

On a fresh install, `this.#tempDir` may not exist, so we should make sure it does

Recursive option here is important

src/extension/survey.ts
```diff
@@ -0,0 +1,150 @@
...
+          // non-zero exit code does not mean failure
+          resolve(e.exitCode)
+        }))
+      })
+    })
+
+    try {
+      const output = readFileSync(tmpfile, { encoding: 'utf-8' }).trim()
+      TelemetryReporter.sendTelemetryEvent('survey.WinDefaultShell', { output })
```
mxsdev, line 128: I think we should report the exit code here too

src/extension/survey.ts
```diff
@@ -0,0 +1,150 @@
...
+      unlinkSync(tmpfile)
+
+      await this.#done()
+    } catch (err) {
+      if (err instanceof Error) {
+        console.log(err.message)
```
mxsdev, line 134: If the command failed, the tempfile may not exist. Then the user will be prompted on every time they open a file. If this is intentional that's fine, just wanted to point out this behavior

mxsdev, line 134: Might want to consider reporting the error here instead of just logging it. I.e., maybe send `{ err }` (maybe just `err.message` since idk if `err` is JSON stringifiable) to `survey.WinDefaultShell`.

mxsdev (changes requested):
Important part is to ensure `this.#tempDir` exists (it didn't appear to exist on my fresh install, resulting in an error)

PR #377, opened 2023-04-04
Match explorer's UX: no leading slashes

sourishkrout:
The original version had leading slashes on all entries. This changes will make the tree view match the explorer view.

image


degrammer (approved):
LGTM

PR #378, opened 2023-04-04
Fix image paths for walkthrough images

christian-bromann:
Using `runme.dev/img/....` image path don't work as file paths can change after every deploy. This patch replaces them using github links.


sourishkrout (approved):
LGTM. Just noticed this issue in edge. 👍

PR #380, opened 2023-04-04
Support multiple terminal windows in Notebook Terminal

mxsdev:
Supports multiple terminal windows in the notebook terminal. Distinguishes an "active" terminal, which is the one that controls PTY rows/cols.

I turned the `examples/README.md` script which gives rows/cols into a watch script. That way, you can see the effect of multiple terminal windows more easily.

The behavior is as follows:

 - If notebook terminal exists, it will use that size by default
   - When you switch to the integrated terminal, it will then switch to the terminal size
   - To switch back to the notebook terminal's sizing, you have to focus it in the notebook (by e.g. left clicking on it)
 - If no notebook terminal exists, it will use integrated terminal sizing
 - If no integrated terminal exists at all (non interactive old output), program should run normally

Note that I have discovered some weird behavior with the watch script and the native integrated terminal. I can file this as a separate issue, since I'm pretty sure it's an unrelated issue.


sourishkrout:
> * If notebook terminal exists, it will use that size by default
>   
>   * When you switch to the integrated terminal, it will then switch to the terminal size
>   * To switch back to the notebook terminal's sizing, you have to focus it in the notebook (by e.g. left clicking on it)
> * If no notebook terminal exists, it will use integrated terminal sizing
> * If no integrated terminal exists at all (non interactive old output), program should run normally

Testing this. Is switching back and forth notebook->integrated and vice versa multiple times is not supported? Is that the expectation? When they are both visible that is.

mxsdev:
@sourishkrout 

Integrated will pretty much be used as long as it is visible.

To switch back to notebook terminal, you currently have to close the integrated, and then focus (left click) the notebook terminal.

If you'd like it to be possible to use notebook terminal dimensions when they are both visible on resize, that's pretty tricky with VSCode's available APIs.

sourishkrout:
> If you'd like it to be possible to use notebook terminal dimensions when they are both visible on resize, that's pretty tricky with VSCode's available APIs.

Right; I remember they don't have the notion of "active" as in "has the focus" in their APIs. Please see my edit to the comment above. Add some more after.

sourishkrout:
Fwiw, tools like tmux & screen do something similar: They don't switch sizes if multiple clients are attached based on activity. It's either MRU or LRU (can't remember the details) based on when they were attached.

sourishkrout:
Unrelatedly, we need to make sure to honor the mime type. E.g. with all opt-ins for terminal true, the PNG/SVG example will trace binary as printable chars in a notebook terminal.

mxsdev:
> If that's hard or unattainable, I'm wondering if a preference of notebook over an integrated size, if a notebook terminal is present is cleaner. The assumption is, that the integrated terminal is always less space-constrained in comparison to a notebook terminal.

Yeah, I see what you're saying. I'll implement that functionality instead 👍 

> Unrelatedly, we need to make sure to honor the mime type. E.g. with all opt-ins for terminal true, the PNG/SVG example will trace binary as printable chars in a notebook terminal.

Separate issue?

sourishkrout:
Sounds good 👍 .

Separate PR for the mime type issue works as long as the anything but `background` remains opt-in.

Great work 👏 . Really loving the notebook terminal.

mxsdev:
@sourishkrout 

Now preferring notebook over integrated

src/client/components/terminal.ts
```diff
@@ -349,18 +349,27 @@ export class TerminalView extends LitElement {
   }
 
   protected firstUpdated(props: PropertyValues): void {
+    console.log('got to first updated cb!!')
```
sourishkrout, line 352: Still required?

sourishkrout (approved):
Works! Please see comment about potential stray console.log.

LGTM 👍 otherwise.

src/client/components/terminal.ts
mxsdev: Oh yeah, not needed.

I need to get better about this haha

mxsdev:
See #387 for issue about mime type

PR #381, opened 2023-04-05
Fix notebook terminal cursor color

mxsdev:
If there is no `--vscode-terminalCursor-foreground` variable, this will revert to the foreground color instead.

This fixes the terminal cursor not appearing in the `Github Light (Colorblind)` theme.


sourishkrout (approved):
👍 I can confirm that I can see the cursor now!

PR #382, opened 2023-04-05
Fix unnecessary error tracking Notebook Terminal input

mxsdev:
I made a mistake when adding this line. 

This PR prevents console from being overwhelmed with error messages on interactive scripts.


degrammer (approved):
LGTM

PR #386, opened 2023-04-06
Parse out trailing dollar signs in cmd sequences

mxsdev:
Executes sequences like `$ echo Hello` as `echo Hello`

Note that this will affect sequences such as:

```
export VAR="
$
"
```

Though usually `$` are escaped in bash so I imagine this kind of a thing to be rather uncommon.


sourishkrout (approved):
👍 works. Let's merge.

PR #389, opened 2023-04-07
Respect custom mime types in Notebook Terminal-enabled cells

### LINKED ISSUE #387: Notebook Terminal does not respect mime type
We're gonna need to rethink the way that mime types are integrated in `src/extensions/executors/runner.ts`.

Currently they are made to only work with the old output system, but now we need to support it in interactive mode.

mxsdev:
Fixes #387


sourishkrout (approved):
👍 LGTM

PR #390, opened 2023-04-07
Allow font customization in Notebook Terminal

mxsdev:
Adds two configurations, `runme.terminal.fontSize` and `runme.terminal.fontFamily`.


degrammer (approved):
Awesome, LGTM

PR #391, opened 2023-04-07
Allow custom rows/vertical resizing in Notebook Terminal

mxsdev:
Allows user to define config `runme.terminal.rows` for the initial row setting for a terminal.

Also adds vertical drag handle, to adjust number of rows.

**Known Issues**:

 - The number of rows resets when re-running a cell. We would need to code a persistence layer to change this
 - Terminal is not "focused" when dragging.

### DIFF
--- a/src/client/components/terminal.ts
+++ b/src/client/components/terminal.ts
@@ -408,21 +408,21 @@ export class TerminalView extends LitElement {
       initialRows: number
     } | undefined
 
-    dragHandle.addEventListener('mousedown', (e) => {
+    const onMouseDown = (e: MouseEvent) => {
       dragState = {
         initialClientY: e.clientY,
         initialRows: this.rows
       }
       e.preventDefault()
       this.terminal?.focus()
-    })
+    }
 
-    window.addEventListener('mouseup', () => {
+    const onMouseUp = () => {
       if (dragState === undefined) { return }
       dragState = undefined
-    })
+    }
 
-    window.addEventListener('mousemove', (e) => {
+    const onMouseMove = (e: MouseEvent) => {
       if (dragState === undefined || !this.fitAddon) { return }
 
       const delta = e.clientY - dragState.initialClientY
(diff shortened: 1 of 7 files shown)


src/client/components/terminal.ts
```diff
@@ -373,6 +394,50 @@ export class TerminalView extends LitElement {
...
+
+      const delta = e.clientY - dragState.initialClientY
+
+      console.log({ delta, cellHeight: this.fitAddon.getCellSize().height })
+
+      const deltaRows = delta / this.fitAddon.getCellSize().height
+      const newRows = Math.round(dragState.initialRows + deltaRows)
+
+      if (newRows !== this.rows) {
+        this.#resizeTerminal(newRows)
```
degrammer, line 434: You've mentioned that resizing the terminal will not focus, but there is any reason `this.terminal?.focus()` won't work here? I just tested with that change, and the terminal focus is kept after resizing.

src/client/components/terminal.ts
```diff
@@ -373,6 +394,50 @@ export class TerminalView extends LitElement {
...
+      if (dragState === undefined) { return }
+      dragState = undefined
+    })
+
+    window.addEventListener('mousemove', (e) => {
+      if (dragState === undefined || !this.fitAddon) { return }
+
+      const delta = e.clientY - dragState.initialClientY
+
+      console.log({ delta, cellHeight: this.fitAddon.getCellSize().height })
```
degrammer, line 428: is this console.log intentional?

src/client/components/terminal.ts
```diff
@@ -373,6 +394,50 @@ export class TerminalView extends LitElement {
...
+        initialRows: this.rows
+      }
+    })
+
+    window.addEventListener('mouseup', () => {
+      if (dragState === undefined) { return }
+      dragState = undefined
+    })
+
+    window.addEventListener('mousemove', (e) => {
```
degrammer, line 423: It's a good idea to remove the event listener at **disconnectedCallback** same applies to **mouseup**

degrammer (changes requested):
This feature is awesome, left some comments

src/client/components/terminal.ts
mxsdev: OH yeah, I definitely should've tried that. Fixed, thanks!

src/client/components/terminal.ts
mxsdev: No, one day I will notice these things before pushing 🥲

src/client/components/terminal.ts
mxsdev: Yep good call 👍

degrammer (approved):
LGTM

PR #402, opened 2023-04-11
Print code on close in background tasks

mxsdev:
image

Prints code on close in notebook terminal for background tasks. Improves UX


degrammer (approved):
LGTM

PR #404, opened 2023-04-12
Use PID for background task

mxsdev:
- When running, "Background Task" becomes "PID: ####" (for background tasks)
   - When clicked on, this opens the task's integrated terminal
 - Removed PID from "Open Terminal" button

Note that this feature requires the latest stateful/runme executable (on main branch, not release)


sourishkrout (approved):
LGTM 👍

PR #423, opened 2023-04-17
Resolve jitter for streaming outputs

sourishkrout:
The first commit in the PR is an observable-less solution whereas with the second commit applied, it will use rxjs.

### DIFF
--- a/examples/vercel/README.md
+++ b/examples/vercel/README.md
@@ -61,7 +61,7 @@ vercel team switch
 
 go on to list projects:
 
-```sh { interactive=false }
+```sh { interactive=true }
 vercel projects list 2>&1 | tail -n +5
 ```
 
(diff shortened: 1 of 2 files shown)


mxsdev (approved):
Observable code looks correct to me 👍 

Fielding @degrammer on whether we want to introduce observables here. I don't have a problem with it, but we might want to reconsider introducing patterns if they ostracize some of our team members.

degrammer:
> Observable code looks correct to me 👍 and correct usage of runner API
> 
> Fielding @degrammer on whether we want to introduce observables here. I don't have a problem with it, but we might want to reconsider introducing patterns if they ostracize some of our team members.

Thank you for quoting me. I don't have any objections; I'm familiar with observables back from my days working with Angular :P

sourishkrout:
> Thank you for quoting me. I don't have any objections; I'm familiar with observables back from my days working with Angular :P

Be careful what you wish for 👻! 🤣

PR #425, opened 2023-04-18
CodeLens support

mxsdev:
Requires stateful/runme#218 to work properly

Still not done:

 - ~~User setting to toggle CodeLens support~~
 - ~~Only apply to markdown files~~
 - ~~Unit tests~~

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -11,16 +11,19 @@ export namespace Serializer {
     metadata?: Metadata
   }
 
+  type TextPosition = number
+
+  export type TextRange = {
+    start: TextPosition
+    end: TextPosition
+  }
+
   export type Cell = {
     metadata?: Metadata
     languageId?: string
     value: string
-    kind: NotebookCellKind.Markup
-  } | {
-    metadata?: Metadata
-    languageId?: string
-    value: string
-    kind: NotebookCellKind.Code
+    textRange?: Serializer.TextRange
+    kind: NotebookCellKind.Markup | NotebookCellKind.Code
   }
 
   export interface Wasm {
--- a/__mocks__/vscode.ts
+++ b/__mocks__/vscode.ts
@@ -31,6 +31,7 @@ export class Uri extends URI {
 
 export const workspace = {
   getConfiguration: vi.fn().mockReturnValue(new Map()),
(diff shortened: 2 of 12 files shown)


sourishkrout:
@mxsdev did you publish the protos to the BSR? This does not seem to work for me.

mxsdev:
@sourishkrout 

My bad! Just pushed latest protos.

Surprised CI passed, I would've expected it to fail on type errors.

sourishkrout:
No sweat. This feature is fantastic!

I suppose the generate types are not used anywhere on the client side? Otherwise, tsc should have blown up. Perhaps something to check into?

mxsdev:
Hmm, the new types are definitely used client side (in `codelens.ts`, we access the `textRange` field, which is a new type)

I'll look into this tomorrow 👍

src/types.ts
```diff
@@ -11,16 +11,19 @@ export namespace Serializer {
     metadata?: Metadata
   }
 
+  type TextPosition = number
+
+  export type TextRange = {
```
sourishkrout, line 16: Since the type's being recreated here and/or where it's being consumed is independent from the protobuf definition makes it so that tsc won't balk.

It'd imagine it's a matter of figuring out how tightly we want to couple against concrete protobuf types. However, it's kinda the point of using type-/codegen to ensure interface compliance using compiler infra.

image

sourishkrout (changes requested):
Implementation works great! Let's consider how we can make the gRPC dependencies more explicit.

src/types.ts
mxsdev: Yeah, the reason I didn't make gRPC types more explicit here is because I was going along with the existing type implementation, which is a remnant of the WASM module. I agree that it should all be based off of gRPC types instead.

How would you feel about splitting this into a separate ticket & PR? Since it affects more than just this part of the code and is otherwise largely independent of codelens support. @sourishkrout

sourishkrout: New ticket works. We might want to make it about removing the WASM serializer interface and standardizing on gRPC. Don't think we have needed to switch back to WASM at all thus far.

sourishkrout:
Good to merge with typing clean up as separate effort 👍 .

PR #426, opened 2023-04-19
Fix OpenVSX publishing

degrammer:
Fixing OpenVSX publishing for platform-specific targets the format is namespace.extension-version@target.vsix.


sourishkrout (commented):
LGTM 👍 . Trying it now.

PR #445, opened 2023-04-24
Convert EOL in non-interactive runner sessions

### LINKED ISSUE #433: Convert EOL in non-interactive scripts
In non-interactive scripts, we should replace `\n` with `\r\n` manually.

Theoretically this shouldn't affect cross-platform, since the behavior of XTermJS is platform-independent. However we should still test it to be sure.

mxsdev:
Fixes #433

### DIFF
--- a/src/extension/runner.ts
+++ b/src/extension/runner.ts
@@ -46,6 +46,7 @@ export interface RunProgramOptions {
   environment?: IRunnerEnvironment
   terminalDimensions?: TerminalDimensions
   background?: boolean
+  convertEol?: boolean
 }
 
 export interface IRunner extends Disposable {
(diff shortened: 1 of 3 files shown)


src/extension/runner.ts
```diff
@@ -681,6 +710,10 @@ export class GrpcRunnerProgramSession implements IRunnerProgramSession {
   get numTerminalWindows() {
     return this.terminalWindows.size
   }
+
+  protected convertEol() {
```
sourishkrout, line 714: I suppose you could use a TS getter here, but it really doesn't matter much.

sourishkrout (approved):
LGTM 👍

Issue #448, opened 2023-04-25
Survey Windows users in CodeLens Run


mxsdev:
Windows users can't use the CodeLens run feature - in its place we should do a survey to gauge interest for such a feature on windows.

PR #449, opened 2023-04-25
Survey Windows users in CodeLens Run 

mxsdev:
Solves #448

### DIFF
--- a/src/extension/survey.ts
+++ b/src/extension/survey.ts
@@ -148,3 +148,57 @@ export class WinDefaultShell implements Disposable {
     this.#disposables.forEach(d => d.dispose())
   }
 }
+
+export class SurveyWinCodeLensRun implements Disposable {
+  static readonly #id: string = 'runme.surveyWinCodeLensRun'
+
+  constructor(
+    protected context: ExtensionContext
+  ) { }
+
+  shouldPrompt() {
+    return isWindows()
+  }
+
+  async prompt(): Promise<void> {
+    if (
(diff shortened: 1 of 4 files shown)


src/extension/survey.ts
```diff
@@ -148,3 +148,57 @@ export class WinDefaultShell implements Disposable {
...
+    await this.#done()
+  }
+
+  async #undo() {
+    await this.context.globalState.update(SurveyWinCodeLensRun.#id, false)
+  }
+
+  async #done() {
+    await this.context.globalState.update(SurveyWinCodeLensRun.#id, true)
+  }
```
christian-bromann, line 201: These don't seem to be used at all 🤔

src/extension/survey.ts
```diff
@@ -148,3 +148,57 @@ export class WinDefaultShell implements Disposable {
...
+      case 'OK': {
+        TelemetryReporter.sendTelemetryEvent('survey.WinCodeLensRun', { })
+      } break
+
+      case 'Dismiss': {
+        return
+      }
+    }
+
+    await this.#done()
```
sourishkrout, line 192: one of them is used here

sourishkrout (approved):
👍 LGTM. I suppose checking for `wsl` is moot since it will show up under `linux` not `win32`.

src/extension/survey.ts
mxsdev: Oh yeah, `#undo()` is not needed anymore I'll eliminate it

PR #450, opened 2023-04-25
Fix non-text outputs

mxsdev:
Non-text outputs are broken due to #445, since we're mangling all non-interactive outputs.

This PR now only does the convert eol functionality on non-interactive cells with type `text/plain`


sourishkrout (approved):
👍 LGTM

PR #453, opened 2023-04-25
Use runme for setting up test pipeline

christian-bromann:
Instead of re-defining our set-up process in CI again, let's use Runme to re-purpose our documentation.

### Note

I currently use the original CLI:

```sh
npx runme run --filename=CONTRIBUTING.md configureNPM
npx runme run --filename=CONTRIBUTING.md setup
npx runme run --filename=CONTRIBUTING.md build
npx runme run --filename=CONTRIBUTING.md test
```

instead of the experimental one:

```sh
npx runme2 configureNPM setup build test
```

Let me know if you prefer the experimental one instead.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
-      - uses: coactions/setup-xvfb@v1
+      - name: 🧪 Setup and Test with Runme
+        uses: coactions/setup-xvfb@v1
         with:
           run: |
             npm i -g runme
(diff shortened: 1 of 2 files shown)


sourishkrout (approved):
👍 LGTM.

As discussed before, there is only one CLI interface. We don't want to use any prototypes anywhere public. They are for internal experimentation purposes only.

sourishkrout:
@christian-bromann, on a separate note: have we considered running this as matrix build? E.g. run it on macOS & Linux? Perhaps even Windows?

christian-bromann:
> have we considered running this as matrix build? E.g. run it on macOS & Linux? Perhaps even Windows?

Ideally it would be great to increase coverage on these platforms. Let's create a separate ticket for it.

Issue #430, opened 2023-04-20
UX Punchlist


sourishkrout:
- [x] Reconsider showing notifications for Copy, Stop Task, etc - are those really noteworthy enough?
- [x] “Run with Runme” button in Explorer context menu
- [x] Add bundled binary’s path to extension’s [EnvironmentVariableCollection](https://code.visualstudio.com/api/references/vscode-api#EnvironmentVariableCollection) so that it’s included in the PATH
- Clean up Settings:
    - [x] Remove changing core defaults. Users should use markdown annotations instead, changing underlying core defaults individually will lead to non-determinism working on shared markdowns
    - [x] Port for Runme Server? Perhaps this should be “Endpoint Port” and only be a client connection setting in case server runs out-of -band
- [x] Consider adding an annotation to skip ENV VAR prompting

For later:

- [ ] Consider “x” button inside of each notebook UX outputs
- [ ] Show "exit" with error code on non-background tasks too

mxsdev:
> Add bundled binary’s path to extension’s [EnvironmentVariableCollection](https://code.visualstudio.com/api/references/vscode-api#EnvironmentVariableCollection) so that it’s included in the PATH

According to my tests, VSCode prepends the shell $PATH to the integrated $PATH. What this means is that, while you can provide access to new executables through `EnvironmentVariableCollection`, you cannot override already provided commands. Unless there is something I'm overlooking here.

Similar issue occurs for spawning a new terminal.

@sourishkrout

PR #459, opened 2023-04-27
Replace 'port' setting with 'custom address'

mxsdev:
Part of #430

### DIFF
--- a/src/extension/extension.ts
+++ b/src/extension/extension.ts
@@ -67,8 +67,10 @@ export class RunmeExtension {
         return window.showErrorMessage(`Failed to start Runme server. Reason: ${e.message}`)
       }
       TelemetryReporter.sendTelemetryErrorEvent('extension.server', { data: (e as Error).message })
-      // eslint-disable-next-line max-len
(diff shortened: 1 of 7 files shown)


src/extension/extension.ts
```diff
@@ -61,7 +61,8 @@ export class RunmeExtension {
         return window.showErrorMessage(`Failed to start Runme server. Reason: ${e.message}`)
       }
       TelemetryReporter.sendTelemetryErrorEvent('extension.server', { data: (e as Error).message })
-      return window.showErrorMessage('Failed to start Runme server, please try to reload the window')
+      // eslint-disable-next-line max-len
+      return window.showErrorMessage(`Failed to start Runme server, please try to reload the window. Reason: ${(e as any).message}`)
```
christian-bromann, line 65: Would prefer not keeping eslint disable around. Happy to bump the `max-len` as well if it is set too strict.
```suggestion
      return window.showErrorMessage(
            'Failed to start Runme server, please try to reload the window. ' +
            `Reason: ${(e as any).message}`
      )
```

christian-bromann (approved):
One comment otherwise, LGTM :shipit:

PR #473, opened 2023-05-02
Update cell names on save

### LINKED ISSUE #461: Code block name gets out of sync when edited
Please take a look at the following scenario. One of the example code cells looks like this:

image

Let's edit it according to:

image

Now when you hit the "CLI button" the default name will be out of sync between extension (deserialize time) vs CLI (real-time):

image

mxsdev:
Fixes #461 

In the future we should probably return a more detailed response from the serialize request, and then use that to update the metadata.

### DIFF
--- a/src/extension/kernel.ts
+++ b/src/extension/kernel.ts
@@ -11,7 +11,7 @@ import {
   NotebookEdit,
   NotebookDocument,
   env,
-  Uri
+  Uri,
 } from 'vscode'
 import { TelemetryReporter } from 'vscode-telemetry'
 
@@ -28,6 +28,7 @@ import { IRunner, IRunnerEnvironment } from './runner'
 import { executeRunner } from './executors/runner'
 import { ITerminalState, NotebookTerminalType } from './terminal/terminalState'
 import { NotebookCellManager, NotebookCellOutputManager, RunmeNotebookCellExecution } from './cell'
+import { SerializerBase } from './serializer'
 
 enum ConfirmationItems {
   Yes = 'Yes',
@@ -55,8 +56,10 @@ export class Kernel implements Disposable {
 
   protected cellManager = new NotebookCellManager(this.#controller)
 
+  protected serializer?: SerializerBase
+
   constructor(
-    protected context: ExtensionContext,
+    protected context: ExtensionContext
   ) {
     const config = workspace.getConfiguration('runme.experiments')
     this.#experiments.set('grpcSerializer', config.get<boolean>('grpcSerializer', true))
@@ -92,6 +95,10 @@ export class Kernel implements Disposable {
     this.runnerReadyListener?.dispose()
   }
 
+  registerSerializer(serializer: SerializerBase) {
+    this.serializer = serializer
+  }
+
   async getTerminalState(cell: NotebookCell): Promise<ITerminalState|undefined> {
     return (await this.getCellOutputs(cell)).getCellTerminalState()
   }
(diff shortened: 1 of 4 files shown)


src/extension/serializer.ts
```diff
@@ -80,6 +86,33 @@ export abstract class SerializerBase implements NotebookSerializer, Disposable {
     })
   }
 
+  protected async handleNotebookSaved({ uri, cellAt }: NotebookDocument) {
+    // update changes in metadata
+    const bytes = await workspace.fs.readFile(uri)
+    const deserialized = await this.deserializeNotebook(bytes, new CancellationTokenSource().token)
+
+    const notebookEdits = deserialized.cells.flatMap((updatedCell, i) => {
+      const updatedName: string|undefined = (updatedCell.metadata as Serializer.Metadata|undefined)?.['runme.dev/name']
```
christian-bromann, line 95: Is this needed?
```suggestion
      const updatedName = (updatedCell.metadata as Serializer.Metadata|undefined)?.['runme.dev/name']
```

src/extension/serializer.ts
```diff
@@ -80,6 +86,33 @@ export abstract class SerializerBase implements NotebookSerializer, Disposable {
...
+
+        return [
+            NotebookEdit.updateCellMetadata(i, {
+            ...oldCell.metadata || {},
+            'runme.dev/name': updatedName,
+          } as Serializer.Metadata)
+        ]
+      }
+
+      return []
```
christian-bromann, line 107: let's return early
```suggestion
      if (!updatedName) {
        return []
      }

      const oldCell = cellAt(i)
      return [
        NotebookEdit.updateCellMetadata(i, {
          ...oldCell.metadata || {},
          'runme.dev/name': updatedName,
        } as Serializer.Metadata)
      ]
```

src/extension/kernel.ts
```diff
@@ -55,8 +56,10 @@ export class Kernel implements Disposable {
 
   protected cellManager = new NotebookCellManager(this.#controller)
 
+  protected serializer?: SerializerBase
```
christian-bromann, line 59: Where do we use this property in the Kernel?

christian-bromann (commented):
Some comments

src/extension/kernel.ts
mxsdev: Good catch, this is an artifact of an old implementation.

christian-bromann (commented):
I think what is left is removing the unit tests from `tests/extension/serializer.test.ts` and add one for `handleNotebookSaved`

mxsdev:
> I think what is left is removing the unit tests from tests/extension/serializer.test.ts

Why remove existing unit tests? Are they outdated? If so, maybe that should be in a separate PR

> and add one for handleNotebookSaved

There is not one already for `handleNotebookChanged`, but I'll still do it for sure 👍

christian-bromann:
> Why remove existing unit tests? Are they outdated? If so, maybe that should be in a separate PR

No, it is just that no changes were done in the `WasmSerializer` class, yet we update the tests for it. Are there needed as part of the PR?

mxsdev:
> No, it is just that no changes were done in the WasmSerializer class, yet we update the tests for it. Are there needed as part of the PR?

Those changes are needed to fix outstanding type errors in `serializer.test.ts`.

christian-bromann:
> Those changes are needed to fix outstanding type errors in `serializer.test.ts`. I guess I can keep the type errors?

Ah, I didn't see that, keep the changes then. Thanks for clarifying.

christian-bromann (approved):
LGTM 👍

Issue #203, opened 2023-02-14
Unit test code cells always being assigned UUIDs


sourishkrout:
This is relevant for adding code cells from the UI as well as deserialization.

Let's make sure we have sufficient test coverage.

Follow up to https://github.com/stateful/vscode-runme/pull/201.

christian-bromann:
@sourishkrout do you think we can close this, given #475 was merged?

sourishkrout:
yes, go for it please

PR #475, opened 2023-05-02
Use cell uuid for terminal identification

### LINKED ISSUE #434: Use cell uuid for terminal identification
We should use cell uuid instead of cell index for terminal identification. 

This prevents weird issues, like The "Open Terminal" button appearing on Markdown cells (reported over slack by @sourishkrout)

mxsdev:
Fixes #434

### DIFF
--- a/src/extension/utils.ts
+++ b/src/extension/utils.ts
@@ -99,9 +99,19 @@ export function getTerminalRunmeId(t: vscode.Terminal): string|undefined {
     ?? undefined
 }
 
+export function getCellRunmeId(cell: vscode.NotebookCell) {
+  return getCellUUID(cell)
+}
+
+function getCellUUID(cell: vscode.NotebookCell): string {
+  return getAnnotations(cell)['runme.dev/uuid']!
+}
+
 export function getTerminalByCell(cell: vscode.NotebookCell): RunmeTerminal | undefined {
+  const RUNME_ID = getCellRunmeId(cell)
+
   return vscode.window.terminals.find((t) => {
-    return getTerminalRunmeId(t) === `${cell.document.fileName}:${cell.index}`
+    return getTerminalRunmeId(t) === RUNME_ID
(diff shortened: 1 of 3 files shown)


src/extension/utils.ts
```diff
@@ -99,9 +99,19 @@ export function getTerminalRunmeId(t: vscode.Terminal): string|undefined {
     ?? undefined
 }
 
+export function getCellRunmeId(cell: vscode.NotebookCell) {
+  return getCellUUID(cell)
+}
+
+function getCellUUID(cell: vscode.NotebookCell): string {
+  return getAnnotations(cell)['runme.dev/uuid']!
+}
```
christian-bromann, line 108: Why do we use two functions here that call each other?

christian-bromann (commented):
One question on the implementation.

Another question: is it always guaranteed that every cell payload comes with a `runme.dev/uuid` property?

src/extension/utils.ts
mxsdev: Because the runme id may change in the future.

For example, we may want to prefix it with the filename, like `${filename}:${uuid}`

mxsdev:
@christian-bromann 

> Another question: is it always guaranteed that every cell payload comes with a runme.dev/uuid property?

For code cells yes, see #203. I guess we could throw an error for non-code cells?

christian-bromann:
> I guess we could throw an error for non-code cells?

Can you elaborate what are non code cells?

mxsdev:
> Can you elaborate what are non code cells?

Cells that are non-runnable, i.e. Markdown cells.

christian-bromann:
> Cells that are non-runnable, i.e. Markdown cells.

Make sense 👍

christian-bromann:
> I guess we could throw an error for non-code cells?

I think this is a good idea so we can detect unintentional use of this method early.

christian-bromann (approved):
LGTM 👍

PR #494, opened 2023-05-09
GitHub authentication

degrammer:
This PR register a command to authenticate with GitHub (used for testing purposes) and a new GitHub service that provides access to GitHub API via Octokit.

The command was used to test a workflow dispatch event https://github.com/stateful/runme-canary/actions/runs/4928319668/jobs/8806523478

how to dispatch a workflow with the API:

1. Get a GitHub Access Token 
```
const session = await authentication.getSession('github', ['repo'], { createIfNone: true })
const githubService = new GitHubService(session.accessToken)
```
2.  Dispatch a workflow
```
await githubService.createWorkflowDispatch({
   owner: 'stateful',
    repo: 'runme-canary',
    workflow_id: 'test-inputs.yml',
    ref: 'main',
    inputs: {
      publishMarketplace: 'no',
      publishOpenVSX: 'no',
      releaseChannel: 'edge',
      releaseType: 'patch',
      releaseVersion: 'v5.0.1'
    }
})
```
This PR also introduces a method to download the YAML file contents from a Workflow file.

The above methods can be called from a cell execution (not included in this PR)

### DIFF
--- a/src/extension/services/github.ts
+++ b/src/extension/services/github.ts
@@ -1,0 +1,42 @@
+import { Octokit } from 'octokit'
+import { parse } from 'yaml'
+
+import { IWorkflowDispatchOptions, IWorkflowYamlContentRequest } from './types'
+
+export class GitHubService {
+    private octokit: Octokit
+    constructor(accessToken: string) {
+        this.octokit = new Octokit({ auth: accessToken })
+    }
+
+    /**
+     *  Dispatch a workflow
+     */
+    async createWorkflowDispatch(dispatchOptions: IWorkflowDispatchOptions):Promise<void> {
+        const { owner, repo, workflow_id, ref, inputs } = dispatchOptions
+        await this.octokit.rest.actions.createWorkflowDispatch({
+            owner,
+            repo,
+            workflow_id,
+            ref,
+            inputs
+        })
+
+    }
+
+    /**
+     * Download the YAML content from a workflow file
+     * @returns A workflow YAML file in JSON format
+     */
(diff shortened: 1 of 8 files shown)


src/extension/services/github.ts
```diff
@@ -0,0 +1,42 @@
...
+     */
+    async getWorkflowYamlFile(options: IWorkflowYamlContentRequest): Promise<string> {
+        const { owner, repo, name } = options
+        const workflow = await this.octokit.rest.repos.getContent({
+            owner,
+            repo,
+            path: `.github/workflows/${name}.yml`
+        })
+
+        const decodedContent = Buffer.from((workflow.data as any).content, 'base64').toString()
```
christian-bromann, line 39: Let's handle the error here:

```suggestion
        if (!('content' in workflow.data)) {
          throw new Error('Failed requesting repository content')
        }

        const decodedContent = Buffer.from(workflow.data.content, 'base64').toString()
```

christian-bromann (commented):
Would be great to get some unit tests on `GitHubService`

src/extension/services/github.ts
degrammer: Fixed!

degrammer:
> Looks reasonable and like we can iterate on it easily in the future 👍
> 
> Seconding what christian said, it would be nice to have unit tests on the `GithubService` class.

Thanks @mxsdev, the idea is to have this PR merged and iterate over this, since I don't want to end up with a huge PR. I agree, tests are coming once I have all the implementation details.

mxsdev (approved):

PR #495, opened 2023-05-09
Parse dotenv files

mxsdev:
Gives the user configuration to set dotenv file schema, and to toggle it in general.

Part of stateful/runme#234

Note that environment variables set in dotenv files "persist," so even if you delete it from a dotenv file, it will still exist in the runner session. Fixing this will require changes to the behavior and/or API of the grpc server.

### DIFF
--- a/src/extension/utils.ts
+++ b/src/extension/utils.ts
@@ -12,14 +12,16 @@ import vscode, {
   NotebookCellExecution,
   NotebookCellOutput,
   commands,
+  WorkspaceFolder,
 } from 'vscode'
 import { v5 as uuidv5 } from 'uuid'
 import getPort from 'get-port'
+import dotenv from 'dotenv'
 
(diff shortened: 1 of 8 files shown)


src/extension/utils.ts
```diff
@@ -360,3 +362,57 @@ export function isWindows(): boolean {
...
+
+  const envFiles = getEnvWorkspaceFileOrder()
+  console.log(envFiles)
+
+  const envs = await Promise.all(
+    envFiles.map(async (fileName) => {
+      const dotEnvFile = Uri.joinPath(workspaceFolder.uri, fileName)
+
+      // console.log(dotEnvFile.fsPath)
+
```
christian-bromann, line 402:
```suggestion
```

christian-bromann (commented):
Just found some development leftovers there..also haven't seen a do..while in a while, nice 😉

degrammer (approved):
Nice, lgtm 🚀

PR #502, opened 2023-05-12
Match subset of words

sourishkrout:
Works locally. Want to see if this passes CI. Open to other suggestions.

### DIFF
--- a/tests/e2e/specs/basic.e2e.ts
+++ b/tests/e2e/specs/basic.e2e.ts
@@ -354,5 +354,5 @@ async function tryExecuteCommand(workbench: Workbench, command: string) {
 // }
 
 async function clearAllOutputs(workbench: Workbench) {
-  await tryExecuteCommand(workbench, 'notebook: clear all outputs')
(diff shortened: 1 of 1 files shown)


tests/e2e/specs/basic.e2e.ts
```diff
@@ -354,5 +354,5 @@ async function tryExecuteCommand(workbench: Workbench, command: string) {
 // }
 
 async function clearAllOutputs(workbench: Workbench) {
-  await tryExecuteCommand(workbench, 'notebook: clear all outputs')
+  await tryExecuteCommand(workbench, 'Notebook: Clear Outputs')
```
mxsdev, line 357:
```suggestion
  await tryExecuteCommand(workbench, 'Notebook: Clear Cell Outputs')
  await tryExecuteCommand(workbench, 'Notebook: Clear All Outputs')
```

I tested this locally, and noticed that (at least for me) some cell outputs lingered.

What do you think about the suggested change?

mxsdev (changes requested):

PR #514, opened 2023-05-15
GitHub Action Cell Renderer

degrammer:
Run a GitHub Action from the Notebook,
You can test it by running the following action: https://github.com/stateful/runme-canary/blob/main/.github/workflows/test-inputs.yml

https://github.com/stateful/vscode-runme/assets/4001529/e4ad4e88-dbb9-4ce9-85dc-d09fc10e3b3a


src/extension/kernel.ts
```diff
@@ -209,6 +210,26 @@ export class Kernel implements Disposable {
...
+          owner: message.output.owner,
+          repo: message.output.repo,
+          run_id: Number(workflowRun.id),
+          onStatusUpdate: (workflowRun) => {
+            postClientMessage(this.messaging, ClientMessages.githubWorkflowStatusUpdate, {
+              workflowRun
+            })
+          }
+        })
+      }
```
christian-bromann, line 232: Maybe we can return early here and safe indentation
```suggestion
      if (!workflowRun) {
        return
      }
      await checkWorkflowRunStatus({
        owner: message.output.owner,
        repo: message.output.repo,
        run_id: Number(workflowRun.id),
        onStatusUpdate: (workflowRun) => {
          postClientMessage(this.messaging, ClientMessages.githubWorkflowStatusUpdate, {
            workflowRun
          })
        }
      })
```

src/client/components/github.ts
```diff
@@ -0,0 +1,434 @@
...
+  @property({ type: Object })
+  state: GitHubState = {}
+  
+  protected disposables: Disposable[] = []
+
+  private inputs: Record<string, string> = {}
+
+  // Define scoped styles right with your component, in plain CSS
+  /* eslint-disable max-len */
+  static styles = css`
```
sourishkrout, line 56: I'd suggest moving this to a `styles.ts` or similar file ahead of a full review @degrammer.

christian-bromann (commented):
Nice feature 🎉  maybe we could consider having an e2e test for this?

sourishkrout (changes requested):
Is it possible to also support dropping the URL to the workflow UI into the cell, eg https://github.com/stateful/runme-canary/actions/workflows/test-inputs.yml ?

I'm more inclined to associate it with a UX rather than the workflow definition. Obviously it'd be great to support both.

sourishkrout:
Also, we do have a parameter-less workflow here: https://github.com/stateful/app/blob/main/.github/workflows/deploy.yml (private repo) which gives me:
```
Unsupported GitHub Workflow, please ensure you are specifying an action with workflow_dispatch
```

However, `workflow_dispatch` is specified. Just no params which I think should still allow us to trigger it, no?

degrammer:
> Also, we do have a parameter-less workflow here: https://github.com/stateful/app/blob/main/.github/workflows/deploy.yml (private repo) which gives me:
> 
> ```
> Unsupported GitHub Workflow, please ensure you are specifying an action with workflow_dispatch
> ```
> 
> However, `workflow_dispatch` is specified. Just no params which I think should still allow us to trigger it, no?

Good point, yes, it should be supported!

sourishkrout:
> Nice feature 🎉 maybe we could consider having an e2e test for this?

Btw, I'm all for adding more e2e tests. However, let's do the very last thing. I'd rather have the UX somewhat final, knowing how time-consuming it can be to get e2e tests to run reliably.

src/client/components/github.ts
degrammer: Moved to a css file and added a webpack raw-loader, it provides better experience for editing CSS

src/extension/kernel.ts
degrammer: Fixed

src/extension/utils.ts
```diff
@@ -123,10 +123,19 @@ export function isDenoScript(runningCell: vscode.TextDocument) {
   return text.indexOf('deployctl deploy') > -1
 }
 
+export function isGitHubLink(runningCell: vscode.TextDocument) {
+  const text = runningCell.getText()
+  return text.includes('.github/workflows')
+
```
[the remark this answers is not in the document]
degrammer: Added this improvement, but I think there are more scenarios to cover here, I'm reviewing it.

degrammer:
> Is it possible to also support dropping the URL to the workflow UI into the cell, eg https://github.com/stateful/runme-canary/actions/workflows/test-inputs.yml ?
> 
> I'm more inclined to associate it with a UX rather than the workflow definition. Obviously it'd be great to support both.
@sourishkrout  Added support, please test again and let me know the results

degrammer:
> Is it possible to also support dropping the URL to the workflow UI into the cell, eg https://github.com/stateful/runme-canary/actions/workflows/test-inputs.yml ?
> 
> I'm more inclined to associate it with a UX rather than the workflow definition. Obviously it'd be great to support both.

Done, added support

degrammer:
> Also, we do have a parameter-less workflow here: https://github.com/stateful/app/blob/main/.github/workflows/deploy.yml (private repo) which gives me:
> 
> ```
> Unsupported GitHub Workflow, please ensure you are specifying an action with workflow_dispatch
> ```
> 
> However, `workflow_dispatch` is specified. Just no params which I think should still allow us to trigger it, no?

Added support for this workflow

__mocks__/vscode.ts
```diff
@@ -255,3 +256,18 @@ export class CodeLens {
...
+
+export class WorkspaceEdit {
+  set() {
+    
+  }
+}
+
+export class NotebookEdit {
+  static updateCellMetadata = vi.fn()
+
```
christian-bromann, line 272:
```suggestion
```

__mocks__/vscode.ts
```diff
@@ -255,3 +256,18 @@ export class CodeLens {
...
 }
+
+export const authentication = {
+  getSession: vi.fn()
+}
+
+export class WorkspaceEdit {
+  set() {
+    
+  }
```
christian-bromann, line 267:
```suggestion
  set() {}
```

src/client/components/github/workflowRun.ts
```diff
@@ -0,0 +1,41 @@
...
+    @property()
+    displayTitle?: string
+
+    @property()
+    avatarUrl?: string
+
+    @property()
+    githubUserName?: string
+
+    static styles = unsafeCSS(require('!!raw-loader!./styles/workflowRun.css').default)
```
christian-bromann, line 25: I don't feel very confident in this approach. The raw-loader project hasn't seen any updates for 3 years and using `require` in an ESM module (as loaded in a browser environment) does not seem to be the right approach.

An alternative on top of my mind is creating e.g. a `workflow.css.ts` file that does something like this:

```ts
export default /*css*/`
  .className { ... }
  // ...
```

which would continue offer us code highlighting and autocompletion but also allow us to simply import the CSS as a string.

Another alternative is to use https://webpack.js.org/loaders/css-loader/.

What do you think?

src/extension/messages/github.ts
```diff
@@ -0,0 +1,39 @@
...
+import { postClientMessage } from '../../utils/messaging'
+import { checkWorkflowRunStatus, deployWorkflow } from '../executors/github/workflows'
+
+export interface IGitHubMessaging {
+    messaging: NotebookRendererMessaging
+    message: ClientMessage<ClientMessages>
+}
+
+export default async function handleGitHubMessage({ messaging, message }: IGitHubMessaging): Promise<void> {
+    if (message.type === ClientMessages.githubWorkflowDispatch) {
```
christian-bromann, line 14: return early and safe indentations, e.g.:
```suggestion
    if (message.type !== ClientMessages.githubWorkflowDispatch) {
        return
    }
```

sourishkrout (changes requested):
Works and looks great! I canceled my workflow half-way through but the UI displayed it as successfully completed. Something we need to fix please.

src/client/components/github/workflowRun.ts
degrammer: Fixed, for now loading the CSS AS string template literal

src/extension/messages/github.ts
degrammer: Done

__mocks__/vscode.ts
degrammer: > Works and looks great!
> 
> 1. I canceled my workflow half-way through but the UI displayed it as successfully completed. Something we need to fix please.
> 2. Any idea why the file continuously shows the unsaved dot, even after saving?
> 
> image

@sourishkrout :
1. Fixed
2. I'm checking this issue, let me know if this is a blocker for merging this.

package.json
```diff
@@ -539,6 +540,7 @@
     "fork-ts-checker-webpack-plugin": "^8.0.0",
     "husky": "^8.0.3",
     "npm-run-all": "^4.1.5",
+    "raw-loader": "^4.0.2",
```
christian-bromann: Do we need this still?

christian-bromann (approved):
LGTM 👍

__mocks__/vscode.ts
sourishkrout: @degrammer re 2., I think we need to understand what the problem is. I'm uncomfortable merging since it feels like a "bug in the making" to me. Sorry about the delay, I overlooked this ping in the noise of everything else.

src/extension/executors/github.ts
```diff
@@ -0,0 +1,35 @@
...
+  this: Kernel,
+  exec: NotebookCellExecution,
+  doc: TextDocument,
+  outputs: NotebookCellOutputManager,
+): Promise<boolean> {
+  try {
+    await authentication.getSession('github', ['repo'], { createIfNone: true })
+    const { owner, repo, path, ref } = parseGitHubURL(doc.getText())
+    const json = await getYamlFileContents({ owner, repo, path })
+    await updateCellMetadata(exec.cell, {
```
sourishkrout, line 19: > @degrammer re 2., I think we need to understand what the problem is. I'm uncomfortable merging since it feels like a "bug in the making" to me. Sorry about the delay, I overlooked this ping in the noise of everything else.

I'm fairly certain this is the problem, @degrammer. The `updateCellMetadata` mutates the *input* cell where this metadata should be associated with the github-specific output generated to trigger the renderer. We might want to rename this method to include `Input`. It shouldn't be called unless the source markdown document requires modification.

src/extension/executors/github.ts
```diff
@@ -0,0 +1,35 @@
...
+import { parseGitHubURL, getYamlFileContents } from './github/workflows'
+
+export async function github(
+  this: Kernel,
+  exec: NotebookCellExecution,
+  doc: TextDocument,
+  outputs: NotebookCellOutputManager,
+): Promise<boolean> {
+  try {
+    await authentication.getSession('github', ['repo'], { createIfNone: true })
```
sourishkrout, line 16: btw, `'github'` here should probably be a constant someplace.

degrammer: fixed

src/extension/cell.ts
```diff
@@ -126,7 +134,7 @@ export class NotebookCellOutputManager {
       case OutputType.vercel: {
         const json: CellOutputPayload<OutputType.vercel> = {
           type: OutputType.vercel,
-          output: metadata['runme.dev/vercelState'] ?? {
+          output: this.cellState?.state as VercelState ?? {
```
mxsdev, line 137: Implementation of `ICellState` is fine, except for one thing:

If the type of a cell changes, say, from `GithubState` to `DenoState`, then the existing state object could get carried over, resulting in the deno object being passed nonsense state.

We could fix this as follows: instead of accessing `this.cellState` directly, we use a `this.getCellState(type: OutputType)` method. When the existing cell state is not of the given type, the cell state could be reset.

degrammer: Thanks @mxsdev, added a new method as recommended

sourishkrout (approved):
Can confirm the feature now works as expected, including the document no longer being marked unsaved.

mxsdev (approved):

PR #518, opened 2023-05-16
Improvements on Windows setup

### LINKED ISSUE #311: `download:binary` script does not work on Windows
One option is to use [`cross-env`](https://www.npmjs.com/package/cross-env).

But honestly I'd rather us not even use that. I feel like we should just not have environment variables at all in `download:binary` or `prepare-binary`, and just detect them within the relevant scripts.

cc @degrammer

christian-bromann:
fixes #311

### DIFF
--- a/.github/scripts/downloadAndExtract.js
+++ b/.github/scripts/downloadAndExtract.js
@@ -22,7 +22,7 @@ const SCHEMA = {
   type: {
     type: 'string',
     short: 'f',
-    default: 'tar',
+    default: process.env.TYPE,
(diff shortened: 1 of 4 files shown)


christian-bromann:
@degrammer would love to get your review on this. I moved the environment variable out of the call and it works now for me on my Windows VM.

mxsdev (approved):
Curious to get @degrammer 's thoughts as well

degrammer (approved):
LGTM 🚀

PR #521, opened 2023-05-18
Runme logging in output channel

### LINKED ISSUE #454: Send logs to VS Code output channel
As part of an initiative to make logging better, we should probably introduce a custom "Runme" output channel to send our logs, rather than doing `console.log`.

This has already affected one user (see #447)

christian-bromann:
fixes #454

This is a simple patch that makes Runme log into an dedicated output channel. I kept it intentionally simple for now as we don't know yet many requirements for our logging. For now, this gives us a central point/module to handle logging. I looked into existing logging package which all seem overkill and came with some issues, e.g. [@vscode-logging/logger](https://github.com/SAP/vscode-logging/blob/master/packages/logger) would print out a JSON blog rather than a simple log line.

Something I would have liked to address but is not possible at the moment:

- colors in output channel: currently not supported (see https://github.com/microsoft/vscode/issues/571)
- add a command to copy out all logs from our channel to support users debugging issues: also currently not supported https://github.com/microsoft/vscode/issues/65108

@mxsdev would this prevent the error reported in the original issue to happen again?

Demo:
Screenshot 2023-05-17 at 16 59 09

Interface:

```ts
import getLogger from './logger'

const log = getLogger('someScope')
log.info('I am an info message')
log.error('I am an error message')
```

### DIFF
--- a/src/extension/index.ts
+++ b/src/extension/index.ts
@@ -19,6 +19,10 @@ export async function activate (context: ExtensionContext) {
     log.info(`Failed to initialize the extension: ${err.message}`)
   }
 
+  log.error('I am an error message')
+  getLogger('Kernel').info('info message with scope')
+  getLogger('Kernel').error('error message with scope')
+
   TelemetryReporter.sendTelemetryEvent('activate')
 }
 
--- a/src/extension/index.ts
+++ b/src/extension/index.ts
@@ -19,10 +19,6 @@ export async function activate (context: ExtensionContext) {
     log.info(`Failed to initialize the extension: ${err.message}`)
   }
 
-  log.error('I am an error message')
-  getLogger('Kernel').info('info message with scope')
-  getLogger('Kernel').error('error message with scope')
-
   TelemetryReporter.sendTelemetryEvent('activate')
 }
 
--- a/src/extension/index.ts
+++ b/src/extension/index.ts
@@ -16,7 +16,7 @@ export async function activate (context: ExtensionContext) {
     await ext.initialize(context)
     log.info('Extension successfully activated')
   } catch (err: any) {
-    log.info(`Failed to initialize the extension: ${err.message}`)
+    log.error(`Failed to initialize the extension: ${err.message}`)
   }
 
   TelemetryReporter.sendTelemetryEvent('activate')
--- a/src/extension/index.ts
+++ b/src/extension/index.ts
@@ -2,25 +2,27 @@ import type { ExtensionContext } from 'vscode'
 import { TelemetryReporter } from 'vscode-telemetry'
 
 import { RunmeExtension } from './extension'
+import getLogger from './logger'
 
 declare const INSTRUMENTATION_KEY: string
 
 const ext = new RunmeExtension()
+const log = getLogger()
 
 export async function activate (context: ExtensionContext) {
   TelemetryReporter.configure(context, INSTRUMENTATION_KEY)
-  console.log('[Runme] Activating Extension')
+  log.info('Activating Extension')
   try {
     await ext.initialize(context)
-    console.log('[Runme] Extension successfully activated')
+    log.info('Extension successfully activated')
   } catch (err: any) {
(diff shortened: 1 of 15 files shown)


src/extension/logger.ts
```diff
@@ -0,0 +1,45 @@
...
+
+function color (color: keyof typeof colors, text: string) {
+  return `${colors[color]}${text}${colors.reset}`
+}
+
+function log (scope?: string, ...logParams: unknown[]) {
+  const now = new Date()
+  const scopeAddition = scope ? color('yellow', `[${scope}]`) : ''
+  const prefix = util.format(
+    `${color('green' ,'[%s]')} Runme%s:%s`,
```
mxsdev, line 30: I think something like

`[TIMESTAMP] [LEVEL] Runme(scope): [msg]`

Might be more consistent with other log formats I've seen

src/extension/index.ts
```diff
@@ -2,25 +2,27 @@ import type { ExtensionContext } from 'vscode'
...
   TelemetryReporter.configure(context, INSTRUMENTATION_KEY)
-  console.log('[Runme] Activating Extension')
+  log.info('Activating Extension')
   try {
     await ext.initialize(context)
-    console.log('[Runme] Extension successfully activated')
+    log.info('Extension successfully activated')
   } catch (err: any) {
-    console.log(`[Runme] Failed to initialize the extension: ${err.message}`)
+    log.info(`Failed to initialize the extension: ${err.message}`)
```
mxsdev, line 19:
```suggestion
    log.error(`Failed to initialize the extension: ${err.message}`)
```

mxsdev (changes requested):
I've been wanting this for months 🤩 thank you for doing this!

src/extension/logger.ts
```diff
@@ -0,0 +1,45 @@
...
+    scopeAddition,
+    // if first log param starts with error symbol make sure to include it here where colors have an effect
+    (logParams[0] as string).startsWith(colors.red) ? ` ${logParams.shift()}` : ''
+  )
+  console.log(prefix, ...logParams)
+  outputChannel.appendLine([prefix, ...logParams].join(' ').replace(COLOR_REGEXP, ''))
+}
+
+export default function getLogger (scope?: string) {
+  return {
```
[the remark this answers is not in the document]
christian-bromann: added

src/extension/logger.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+function log (scope?: string, ...logParams: unknown[]) {
+  const now = new Date()
+  const scopeAddition = scope ? color('yellow', `[${scope}]`) : ''
+  const prefix = util.format(
+    `${color('green' ,'[%s]')} Runme%s:%s`,
+    now.toISOString(),
+    scopeAddition,
+    // if first log param starts with error symbol make sure to include it here where colors have an effect
```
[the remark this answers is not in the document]
christian-bromann: good call 👍

christian-bromann:
Addressed comments and fixed tests.

mxsdev (approved):
LGTM

christian-bromann:
Screenshot 2023-05-18 at 11 14 23

src/extension/logger.ts
```diff
@@ -0,0 +1,47 @@
+import util from 'node:util'
+
+import { window } from 'vscode'
+
+const outputChannel = window.createOutputChannel('Runme')
+
+const COLOR_REGEXP = /[\u001b\u009b][[()#;?]*(?:[0-9]{1,4}(?:;[0-9]{0,4})*)?[0-9A-ORZcf-nqry=><]/g
```
mxsdev, line 7: We already have a `stripAnsi` function in `src/utils/ansi`

mxsdev (changes requested):
Forgot to mention one last thing!

src/extension/logger.ts
christian-bromann: replaced 👍

Issue #519, opened 2023-05-17
Allow referencing output of previous cell


sourishkrout:
Much like Jupyter's `_` convention to be able to reference the previous cell's output, let's implement something similar.

Pending viability, we could try to use `$__` (since `$_`) is already defined. If possible, we could also consider `$_1`, `$_2`, `$_3`, etc to reference the n-last cell relative to the current one.

However, let's prioritize the simple `$__` first, please.

PR #522, opened 2023-05-18
Store last stdout in `$__`

mxsdev:
Solves #519

Requires stateful/runme#250 (proto changes needed for build)


sourishkrout (approved):
Works when rebased with `main` (new protos) 👍. Let's merge so we can ship please.

Issue #497, opened 2023-05-10
Configuration sometimes not openable before running first script


mxsdev:
Sometimes, clicking on configuration to open the annotation doesn't work until you've run at least one script.

Discovered by @admc

sourishkrout:
> Sometimes, clicking on configuration to open the annotation doesn't work until you've run at least one script.
> 
> Discovered by @admc

Do you have any idea what's causing this? We used to have the issue that the notebook kernel wasn't ready and needed selection. However, I thought we caught that exception and prompted the user accordingly.

mxsdev:
Can't say at the moment, I'd have to do some further investigation to be sure. 

Best guess is it has something to do with not being allowed to create a notebook execution?

sourishkrout:
Makes sense. If it happens again, a screenshot or something would be helpful (cc @admc).

sourishkrout:
I have a hunch that the multiple output refactor rejiggered the code so that the "kernel not selection" exception is no longer caught and handled @mxsdev. I'll try to repro it again later; however, I did see the exception in the logging console (unhandled).

sourishkrout:
Have not seen this error crop up ever since we reshuffled some code. Please open again when you run up against it @mxsdev @admc.

PR #538, opened 2023-05-24
Prompt user to select default kernel

mxsdev:
Should solve #497, though I don't have a repro unfortunately

TODO: tests

### DIFF
--- a/src/extension/cell.ts
+++ b/src/extension/cell.ts
@@ -252,7 +252,7 @@ export class NotebookCellOutputManager {
         'Please select a kernel (top right: "Select Kernel") to continue.')
       return
     }
-    return window.showInformationMessage(
+    return await window.showInformationMessage(
       'Please select a notebook kernel first to continue.',
       'Select Kernel'
     ).then(option => {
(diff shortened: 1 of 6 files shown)


sourishkrout:
Can't repro either, which might be a good thing 🤷 ?

However, I'm seeing a lot of these:

```ERR Cannot get cell UUID for non-code cell!```

sourishkrout:
Ah!

image

Which also explains why the "Stop Task" button disappeared from the status bar.

sourishkrout (changes requested):
Please see comments.

mxsdev:
> ERR Cannot get cell UUID for non-code cell!

This I think is a separate issue. I pushed a fix anyway though

mxsdev:
> Can't repro either, which might be a good thing 🤷 ?

I didn't eliminate the issue, only made it like before, where you would get prompted to select a default kernel.

I don't know where vscode stores the "default kernel" setting, but we would need to reset that to test it properly.

sourishkrout:
Thanks for fixing the unrelated bug.

There's a chance that the kernel selection logic is something the VS Code platform improved. I remember researching it and people saying that this "select kernel first" issue couldn't be avoided in "VS Code API user-land".

sourishkrout (approved):
Let's merge and move on for now 👍 .

PR #543, opened 2023-05-25
Convert textRange bytes to characters

mxsdev:
Fixes stateful/runme#269

The `TextRange` information reported from grpc gives offset in bytes, but VSCode expects offset in characters.

My solution was to count the lines manually with rudimentary string operations. This has comically awful performance, but doing it the "right" way is tough, since we'd need to iterate through characters.

For now it should be fine on small-ish files. But if we run into problems with performance for CodeLens, this is the first thing to optimize. [I found a library too](https://www.npmjs.com/package/codepoint-iterator) that could be of help.

Another thing: anything that's not `utf-8` and has wide characters will almost certainly break. VSCode does not appear to let you determine the encoding of a `TextDocument` - see microsoft/vscode#824 - so we can't really figure out the right encoding. Hopefully `UTF-16` and others are uncommon enough that this doesn't become an issue.

### DIFF
--- a/__mocks__/vscode.ts
+++ b/__mocks__/vscode.ts
@@ -285,3 +285,17 @@ export class WorkspaceEdit {
 export class NotebookEdit {
   static updateCellMetadata = vi.fn()
 }
+
+/**
+ * Represents an end of line character sequence in a {@link TextDocument document}.
+ */
+export enum EndOfLine {
+  /**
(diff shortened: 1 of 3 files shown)


sourishkrout (approved):
Works 👍 . I agree, fixing it is more important in the short term than optimizing for run time which we can still do as we encounter larger files.

PR #545, opened 2023-05-26
Support custom shells w/ frontmatter

mxsdev:
Support custom shells with frontmatter

In the future we should have this as a configuration option as well. Lmk if you want to wrap that up with this pr @sourishkrout 

Depends on #544


sourishkrout (approved):
👍 LGTM

PR #558, opened 2023-05-29
Introducing run by category

degrammer:
Work done with @duvanmonsa


duvanmonsa (approved):

package.json
```diff
@@ -226,6 +246,19 @@
...
+        {
+          "command": "runme.runCategory",
+          "when": "notebookType == runme && notebookHasCategories",
+          "group": "navigation"
+        }
+      ],
+      "notebook/cell/execute": [
+        {
+          "command": "runme.runCellCategory",
+          "group": "navigation"
```
mxsdev, line 260:
```suggestion
          "group": "navigation",
          "when": "notebookType == runme && notebookHasCategories"
```

Need to include the same when clause - otherwise it shows up in other notebook types.

src/client/components/configuration/categorySelector.ts
```diff
@@ -0,0 +1,170 @@
...
+                    readonly
+                    class="annotation-item">
+                    <b>${this.identifier}: </b>${this.description}
+                    </vscode-text-field>`
+                }
+                </div>
+                <div class="row">
+                ${when(this.categories.length, 
+                    () => html`
+                    <vscode-button
```
mxsdev, line 155: image

Can we line these up?

mxsdev (changes requested):
Looks fine to me, other than comments.

One big concern is the overall difficulty in using VSCode's `"notebook.execute"` command, which forces a prompt on every command, and also forces the user to manually scroll to the command to see it play.

In a follow-up PR, we may want to think about implementing our own version of this, which scrolls to each command as it runs.

src/client/components/configuration/categorySelector.ts
mxsdev: image

The padding between the buttons and the input box are also not consistent. Should be solvable with the appropriate flex box settings.

package.json
degrammer: Fixed

degrammer:
> Looks fine to me, other than comments.
> 
> One big concern is the overall difficulty in using VSCode's `"notebook.execute"` command, which forces a prompt on every command, and also forces the user to manually scroll to the command to see it play.
> 
> In a follow-up PR, we may want to think about implementing our own version of this, which scrolls to each command as it runs.

Thanks @mxsdev , yeah, we've found the same when testing the run all feature, will address as a separate PR

mxsdev (approved):

PR #560, opened 2023-05-30
cwd support

mxsdev:
Support cwd in frontmatter, annotations. Implemented identically to as in stateful/runme#270


sourishkrout (approved):
👍 LGTM

PR #564, opened 2023-05-31
Fix frontmatter parsing as `[object Object]`

mxsdev:
Fixes stateful/runme#280

For me, the notebook fails to save in this PR; see #566 for more details.


christian-bromann (approved):
👍

sourishkrout (changes requested):
Let's not merge until we have the "can't save doc problem" fixed. If that means tackling the referenced issue, let's go for it please.

sourishkrout (approved):
Moving ahead with a merge. Loose ends to tie up are document in https://github.com/stateful/vscode-runme/issues/566.

PR #565, opened 2023-05-31
Fix notebook not saving

mxsdev:
The notebook will fail to save in PR #564. This fixes that issue.

Note that this requires new VSCode API's which were note present in `v1.72.0`, so I've bumped the version. Let me know if this is ok @christian-bromann @sourishkrout. I could handle the case where the user is not on a new-enough version of vscode if we think that would be a good idea.


mxsdev:
Realized this doesn't really solve the problem, since we don't know what changed during the pre-save event.

Issue #520, opened 2023-05-17
Possible to prompt for extension to be added to extensions.json


sourishkrout:
`extension.json` aka "recommended extensions". Is there an existing command that brings up the dialog?

1. Let's clarify that question.
2. Then, let's figure out how to "schedule" that dialog thoughtfully.

sourishkrout:
@degrammer next one up for you once you current tasks have wrapped. Ping me if you have questions.

PR #569, opened 2023-06-01
Add Runme to recommended extensions

degrammer:
Prompt the user to add the extension to recommended extensions

Note: It does not cover Multi-root workspace for now.
- It prompts the user on Extension Activation (unless disabled by the user)
- It supports don't ask again feature
- Added as a command in order to bring it back again

🎥 Demo 

https://github.com/stateful/vscode-runme/assets/4001529/f8085ef4-1a51-4c9b-b832-9b05bda15181

Fixes https://github.com/stateful/vscode-runme/issues/520

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -112,7 +112,7 @@ export const LANGUAGES = new Map([
  * For example, the vscode language id for all bash-like shells is
  * "shellscript," so this object maps "shellscript" -> "sh"
  */
-export const VSCODE_LANGUAGEID_MAP: Record<string, string|undefined> = {
+export const VSCODE_LANGUAGEID_MAP: Record<string, string | undefined> = {
   'shellscript': 'sh',
   'javascriptreact': 'jsx',
   'typescriptreact': 'tsx',
@@ -1042,3 +1042,8 @@ export const SUPPORTED_FILE_EXTENSIONS = [
   'sh', 'bash', 'ksh', 'zsh', 'fish', 'bat', 'cmd', 'pwsh',
   ...Object.keys(LANGUAGE_PREFIXES),
 ]
+
+export const EXTENSION_NAME = 'stateful.runme'
+export enum TELEMETRY_EVENTS {
+  RecommendExtension = 'runme.recommendExtension'
+}
--- a/src/extension/extension.ts
+++ b/src/extension/extension.ts
@@ -26,7 +26,8 @@ import {
   openIntegratedTerminal,
   authenticateWithGitHub,
   displayCategoriesSelector,
-  runCellsByCategory
+  runCellsByCategory,
+  addToRecommendedExtensions
 } from './commands'
 import { WasmSerializer, GrpcSerializer } from './serializer'
 import { RunmeLauncherProvider } from './provider/launcher'
@@ -35,6 +36,7 @@ import { GrpcRunner, IRunner } from './runner'
 import { CliProvider } from './provider/cli'
 import * as survey from './survey'
 import { RunmeCodeLensProvider } from './provider/codelens'
+import { MessagingBuilder, RecommendExtensionMessage } from './messaging'
 
 export class RunmeExtension {
   async initialize(context: ExtensionContext) {
@@ -80,6 +82,8 @@ export class RunmeExtension {
     const activeUsersSurvey = new survey.SurveyActiveUserFeedback(context)
     const winCodeLensRunSurvey = new survey.SurveyWinCodeLensRun(context)
     const stopBackgroundTaskProvider = new StopBackgroundTaskProvider()
+    const messageBuilder = new MessagingBuilder([new RecommendExtensionMessage(context)])
+    messageBuilder.activate()
 
     const runCLI = runCLICommand(context.extensionUri, !!grpcRunner, server, kernel)
 
@@ -103,6 +107,7 @@ export class RunmeExtension {
       treeViewer,
       activeUsersSurvey,
       winCodeLensRunSurvey,
+      messageBuilder,
       workspace.registerNotebookSerializer(Kernel.type, serializer, {
         transientOutputs: true,
         transientCellMetadata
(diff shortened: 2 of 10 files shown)


degrammer:
@sourishkrout  in case you want to take a look

src/extension/messaging.ts
```diff
@@ -0,0 +1,120 @@
...
+        for (const message of this.messages) {
+            message.display()
+        }
+    }
+}
+
+
+export class RecommendExtensionMessage extends DisplayableMessage implements Disposable {
+    async display(): Promise<void> {
+        try {
```
christian-bromann, line 45: I wonder if it makes sense to wrap this all into a try/catch block. It could potentially swallow errors that we don't expect.

src/extension/extension.ts
```diff
@@ -147,9 +152,16 @@ export class RunmeExtension {
...
+      window.registerUriHandler(uriHandler),
+
+      /**
+       * Runme Message Display commands
+       */
+      RunmeExtension.registerCommand('runme.addToRecommendedExtensions', () => addToRecommendedExtensions(context))
     )
 
+
+
```
christian-bromann, line 164:
```suggestion
```

src/extension/messaging.ts
```diff
@@ -0,0 +1,120 @@
...
+                const recommendations = [EXTENSION_NAME]
+                await workspace.fs.writeFile(extensionsFile, Buffer.from(JSON.stringify({
+                    recommendations
+                })))
+            } else {
+                extensionRecommendations.push(EXTENSION_NAME)
+                await workspace.fs.writeFile(extensionsFile, Buffer.from(JSON.stringify({
+                    recommendations: extensionRecommendations
+                })))
+            }
```
christian-bromann, line 107:
```suggestion
            const recommendations = extensionfileOrDirectoryExists
                ? extensionRecommendations.push(EXTENSION_NAME) && extensionRecommendations
                : [EXTENSION_NAME]
            await workspace.fs.writeFile(extensionsFile, Buffer.from(JSON.stringify({
                recommendations
            })))
```

sourishkrout (changes requested):
Can the JSON be written back to the file pretty-print? Making it compact will likely make git diffs unreadable.

Also, before merging, let's be sure to disable the auto pop-up. We need to determine what conditions should be met for it to pop-up which we can work on in a separate PR.

sourishkrout:
Below's error pops up when I open a workspace (not just a folder). Since the user didn't request "anything", I don't think we should show an error notif in this case. Logging is sufficient. **Not** supporting workspaces is fine, since it's ambiguous what `extension.json` to add Runme to with multiple repos involved.

image

degrammer:
> Can the JSON be written back to the file pretty-print? Making it compact will likely make git diffs unreadable.
> 
> Also, before merging, let's be sure to disable the auto pop-up. We need to determine what conditions should be met for it to pop-up which we can work on in a separate PR.

@sourishkrout  the popup is not displayed in the following conditions:
- The Runme extension is already added to recommendations.json
- The user opt-out via Don't ask me again

Am I missing any other scenario?

sourishkrout:
> > Can the JSON be written back to the file pretty-print? Making it compact will likely make git diffs unreadable.
> > Also, before merging, let's be sure to disable the auto pop-up. We need to determine what conditions should be met for it to pop-up which we can work on in a separate PR.
> 
> @sourishkrout the popup is not displayed in the following conditions:
> 
> * The Runme extension is already added to recommendations.json
> * The user opt-out via Don't ask me again
> 
> Am I missing any other scenario?

That's a great start. What I mean is that I don't want to display the message to the entire install base. The idea would be only to display it if you have run a cell recently or something like that—some activation events. Not suggesting the current implementation "is wrong".

degrammer:
> 

Fixed, now stores the JSON adequately formatted.

degrammer:
> > > Can the JSON be written back to the file pretty-print? Making it compact will likely make git diffs unreadable.
> > > Also, before merging, let's be sure to disable the auto pop-up. We need to determine what conditions should be met for it to pop-up which we can work on in a separate PR.
> > 
> > 
> > @sourishkrout the popup is not displayed in the following conditions:
> > 
> > * The Runme extension is already added to recommendations.json
> > * The user opt-out via Don't ask me again
> > 
> > Am I missing any other scenario?
> 
> That's a great start. What I mean is that I don't want to display the message to the entire install base. The idea would be only to display it if you have run a cell recently or something like that—some activation events. Not suggesting the current implementation "is wrong".

Good point, let's discuss that then

src/constants.ts
```diff
@@ -119,3 +119,7 @@ export enum AuthenticationProviders {
 }
 export const NOTEBOOK_AVAILABLE_CATEGORIES = 'notebookAvailableCategories'
 export const NOTEBOOK_HAS_CATEGORIES = 'notebookHasCategories'
+export const EXTENSION_NAME = 'stateful.runme'
+export enum TELEMETRY_EVENTS {
+  RecommendExtension = 'runme.recommendExtension'
```
sourishkrout, line 124: When the registration of a command is wrapped with `Extension.registerCommand` telemetry is automatically reported. So I'd suggest to remove any other telemetry reporting code please.

src/extension/messaging.ts
```diff
@@ -0,0 +1,120 @@
...
+    }
+}
+
+
+export class RecommendExtensionMessage extends DisplayableMessage implements Disposable {
+    async display(): Promise<void> {
+        try {
+            const skipPrompSettings = this.stateSettings && this.stateSettings[TELEMETRY_EVENTS.RecommendExtension]
+            let promptUser = skipPrompSettings ||
+                this.context.globalState.get(TELEMETRY_EVENTS.RecommendExtension, true)
```
sourishkrout, line 48: @degrammer, I believe there's a workspaceState or something like that? Should be record the "skipping" globally? Meaning it will never ever be shown again? Irrespective of the repo one's working on 🤔. I am thinking out loud here.

sourishkrout (changes requested):
See my questions please @degrammer

src/extension/messaging.ts
degrammer: @sourishkrout  global state works across repositories. Using workspaceState will do the opposite of what you want.

sourishkrout: Well, what I want - I think - is that if I dismissed the recommendation for a particular repo once that it won't apply to all future repos, I am opening. Perhaps we need a "No, thanks" and a "Never ask again" button?

sourishkrout (approved):
As discussed, let's merge 👍 . Opening a new ticket with remaining items.

degrammer:
Follow up ticket: https://github.com/stateful/vscode-runme/issues/616

PR #573, opened 2023-06-02
Remove transient metadata keys from diffs

sourishkrout:
Avoid confusion while reviewing diffs. It will basically avoid scenarios like these:

image

### DIFF
--- a/src/extension/extension.ts
+++ b/src/extension/extension.ts
@@ -1,6 +1,7 @@
 import { workspace, notebooks, commands, ExtensionContext, tasks, window, Uri } from 'vscode'
 import { TelemetryReporter } from 'vscode-telemetry'
 
+import { Serializer } from '../types'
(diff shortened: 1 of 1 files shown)


sourishkrout:
Please do a retro-review, @mxsdev, merging for now.

mxsdev:
LGTM - didn't even know about this feature.

Only one question: is there a consequence to getting rid of these keys? @sourishkrout 

```
          inputCollapsed: true,
          outputCollapsed: true,
```

sourishkrout:
> LGTM - didn't even know this was a feature.
> 
> Only one question: is there a consequence to getting rid of these keys? @sourishkrout
> 
> ```
>           inputCollapsed: true,
>           outputCollapsed: true,
> ```

Nope, those are attribute names we've never used. You could argue they should have never been there.

PR #585, opened 2023-06-05
Allow to define a start-up file to be opened by Runme when extension activates

christian-bromann:
> As a repository maintainer I want to have developers open an e.g. `CONTRIBUTING.md` file every time they start the project. 

This patch adds a `runme.flags.startFile` setting that allows to define a file to be opened by Runme at application start. This would be helpful for anyone that has an important file for contributor to look at to have it opened when VS Code starts. Especially for developers defining a `devcontainer.json` file it could allow new contributors to see all important documentation and commands to start get productive.

If the file is not existing in the repository, it will just return silently. This allows devs to have a global configuration to always open e.g. a `CONTRIBUTING.md` file if existing. Otherwise this option is meant to be used mainly in `.vscode/settings.json` files.

I am not entirely sure if it is a desired behavior for experienced developers to have the same file over and over opened again when starting their IDE. I could think of adding a confirmation popup asking to open again on start up. Feedback on this would be appreciated.

### DIFF
--- a/__mocks__/vscode-telemetry.ts
+++ b/__mocks__/vscode-telemetry.ts
@@ -1,5 +1,6 @@
 import { vi } from 'vitest'
 
 export const TelemetryReporter = {
-  sendTelemetryEvent: vi.fn()
+  sendTelemetryEvent: vi.fn(),
+  sendTelemetryErrorEvent: vi.fn()
 }
--- a/package.json
+++ b/package.json
@@ -313,6 +313,11 @@
             "default": true,
             "markdownDescription": "If enabled, Runme's save restriction on non checked-in files will be lifted."
           },
+          "runme.flags.startFile": {
+            "type": "string",
+            "scope": "window",
+            "markdownDescription": "Relative path to a file within the workspace to open within a Runme notebook view, when VS Code starts (e.g. `./CONTRIBUTING.md`)."
+          },
           "runme.experiments.grpcSerializer": {
             "type": "boolean",
             "scope": "machine",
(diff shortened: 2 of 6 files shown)


sourishkrout:
> I am aware of the `runOn` feature but doubt it fulfils to provide the same seamless experience we use for the Runme link I was looking for. I will play around with your suggested approach and will report back.

The best match would probably be a CLI command that triggers the task execution inside VS Code. Ideally, the Kernel would absorb task management/visibility so that, no matter where you start a task, CLI, or Extension, it's aware if they are already running. To unblock you we could use the file-based approach but I wouldn't want to document and advocated it broadly (yet).

christian-bromann:
> Ideally, the Kernel would absorb task management/visibility so that, no matter where you start a task, CLI, or Extension, it's aware if they are already running.

Can you explain what you mean by that?

I don't think this feature makes sense anywhere outside of VS Code context which is why this is solely added in the extension. To reiterate: the goal is to allow a user (project contributors) to have the extension automatically open a notebook when the IDE opens. project maintainer can set this pref as part of a `devcontainer.json` or `.vscode/settings.json`. This is not about running any tasks, just opening the right start file in the IDE.

sourishkrout:
> > Ideally, the Kernel would absorb task management/visibility so that, no matter where you start a task, CLI, or Extension, it's aware if they are already running.
> 
> Can you explain what you mean by that?
> 
> I don't think this feature makes sense anywhere outside of VS Code context which is why this is solely added in the extension. To reiterate: the goal is to allow a user (project contributors) to have the extension automatically open a notebook when the IDE opens. project maintainer can set this pref as part of a `devcontainer.json` or `.vscode/settings.json`. This is not about running any tasks, just opening the right start file in the IDE.

Yes, I understand. I got a bit ahead of myself here, which I see makes this confusing. However, I meant having a command that allows opening a file in the notebook view in the Extension from the CLI. Trigger by a attach or launch "hook". Further removing the boundaries between the two clients. Anyways, the implemented approach is good enough to unblock you.

sourishkrout:
@christian-bromann merge?

christian-bromann:
@sourishkrout cheers!

Issue #566, opened 2023-05-31
Need better solution for updating name in response to cell content changes


mxsdev:
In #473, I introduced code that modifies the notebook document on save.

I recently ran into an issue where this makes it impossible to save the document, since the document modification itself makes the document need another save.

I think we need a better solution for updating names based on content changes.

In the short-term, we should probably remove that code, so that people can save their documents properly.

cc @sourishkrout

sourishkrout:
We can probably fix/salvage the implementation using the `onWillSaveTextDocument` event instead. It appears it was introduced for the exact use case where changes need to be handled before saving.

https://github.com/microsoft/vscode/blob/main/src/vscode-dts/vscode.d.ts#L11776-L11796

sourishkrout:
> @sourishkrout
> 
> I tried that, but I was unable to get the unsaved file contents the way you can for a text document. In particular, I couldn't get the associated `NotebookData` with the `NotebookDocument`, which would be necessary for serialization.
> 
> I honestly think the easiest solution is to just re-implement the name generation code from go into javascript - would you be okay with this as a short-term solution?

Let's earmark this to chat about in person this week.

sourishkrout:
@mxsdev this is fixed with the `--index` approach, no?

mxsdev:
@sourishkrout 

It unblocks the issue, but the `--index` approach is still not ideal since the generated command is not very legible or re-usable as the document changes.

PR #593, opened 2023-06-07
Run CLI with index rather than cell name

mxsdev:
Requires stateful/runme#293

CLI button now uses cell index rather than cell name. Note that this feature will break for users who don't upgrade to the latest runme.

This unblocks us on [this issue](https://github.com/stateful/vscode-runme/issues/566#issuecomment-1574128419), though it doesn't solve it.

I also introduced `runme.cli.useIntegratedRunme`, which makes it easier to test the new version of runme (uses integrated path).

### DIFF
--- a/tests/extension/commands/index.test.ts
+++ b/tests/extension/commands/index.test.ts
@@ -1,6 +1,6 @@
 import path from 'node:path'
 
-import { beforeEach, expect, test, vi } from 'vitest'
+import { beforeEach, expect, test, vi, suite } from 'vitest'
 import {
   window,
   env,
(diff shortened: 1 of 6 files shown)


tests/extension/commands/index.test.ts
```diff
@@ -106,21 +111,96 @@ test('copyCellToClipboard', () => {
...
+      isDirty: false
+    }
+
+    // vi.mocked(getAnnotations).mockReturnValueOnce({
+    //   name: 'foo-bar',
+    // } as any)
+
+    // vi.mocked(getBinaryPath).mockReturnValueOnce(Uri.file('/bin/runme'))
+    // vi.mocked(window.showInformationMessage).mockResolvedValueOnce(false as any)
+
```
mxsdev, line 141: Unneeded comments

sourishkrout (approved):
👍 LGTM

PR #595, opened 2023-06-09
Use prettier

mxsdev:
Outfits the repo to use prettier.

Also enables auto-save by default in `settings.json`, I hope this is not controversial.

PR #613, opened 2023-06-14
Support custom cell executions

mxsdev:
Allows running cells with custom languages. Provides new `program` cell annotation, with which the cell will run.

Requires runme with stateful/runme#302

Depends on #590

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -17,6 +17,10 @@ This markdown file contains some custom examples to test the execution within a 
 echo "Hello World!"
 ```
 
+```javascript { name=echo-hello-js }
+console.log('Hello World!')
+```
+
 ## More Shell
 
 ```sh { interactive=false }
@@ -164,6 +168,6 @@ curl -s "https://api.marquee.activecove.com/getWeather?lat=52&lon=10" | fx
 
 ```yaml
 config:
-  netsed:
+  nested:
     para: true
 ```
--- a/examples/README.md
+++ b/examples/README.md
@@ -17,10 +17,6 @@ This markdown file contains some custom examples to test the execution within a 
 echo "Hello World!"
 ```
 
-```javascript { name=echo-hello-js }
-console.log('Hello World!')
-```
-
 ## More Shell
 
 ```sh { interactive=false }
@@ -170,4 +166,10 @@ curl -s "https://api.marquee.activecove.com/getWeather?lat=52&lon=10" | fx
 config:
   nested:
     para: true
+```
+
+## JavaScript
+
+```javascript { name=echo-hello-js }
+console.log('Hello World!')
 ```
--- a/examples/README.md
+++ b/examples/README.md
@@ -128,13 +128,30 @@ You can copy also results from the inline executed shell:
 openssl rand -base64 32
 ```
 
-## Non-Supported Languages
+## Non-Shell Languages
 
-These are shown as simple markdowns, e.g:
+These are sometimes executable by default, like for python:
 
-```py { readonly=true }
-def hello():
-    print("Hello World")
+```py
+print("Hello World")
+```
+
+Otherwise, execution can be set with the `interpreter` annotation, like so:
+
+```yaml { interpreter=cat }
+config:
+  nested:
+    para: true
+```
+
+Non-shell scripts can also access environment variables, and are run from the current working directory:
+
+```sh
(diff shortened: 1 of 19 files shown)


tests/extension/messages/cellOutput.test.ts
```diff
@@ -78,4 +80,4 @@ suite('Handle CellOutput messages', () => {
     test('Vercel cell output', async () => {
         return expectOutput(OutputType.vercel)
     })
-})
```
duvanmonsa, line 81: what is the standard for us having an empty line at the end?

duvanmonsa (approved):
LGTM

tests/extension/messages/cellOutput.test.ts
mxsdev: #595

examples/README.md
```diff
@@ -151,3 +155,11 @@ echo Rows: \$(tput lines)
 echo Columns: \$(tput cols)
 "
 ```
+
+## YAML
+
+```yaml
+config:
+  netsed:
```
sourishkrout, line 163:
```suggestion
  nested:
```

sourishkrout (changes requested):
Let's talk this thru today. I'd like us to provide more clarity that this, in fact, works like a `#!` shebang. We can decide to do that in a follow-up PR (likely more UX than logic).

Outside of that something like `/usr/bin/env python3` does not work and errors will be "swallowed". Can we improve that please. Using `env` as a way to allow multi argument shebangs is commonplace.

mxsdev:
> Let's talk this thru today. I'd like us to provide more clarity that this, in fact, works like a #! shebang. We can decide to do that in a follow-up PR (likely more UX than logic).

Agreed. I used "`program`" as a placeholder (it fits with the internal naming structure), but I presumed we would be working it out more.

> Outside of that something like /usr/bin/env python3 does not work and errors will be "swallowed". Can we improve that please. Using env as a way to allow multi argument shebangs is commonplace.

This is a bit complicated since we need to distinguish the program path from the arguments (this is probably a limitation of the syscall responsible for spawning a new process). Some paths can contain spaces, like "C:\Program Files\...". For now I can use a simple regexp on whitespace.

But eventually we would need to (in descending order of difficulty):

1. Do some basic shell parsing (maybe a use case for shlex)
2. Wrap program execution in a bash shell (makes determining if the program exists [to create error window indicating unsupported language] difficult)
3. Regexp/split on whitespace (breaks filepaths with spaces)
4. Provide a separate input field for arguments (if i recall correctly, arguments are parsed by the system/shell program, so i don't think we need to worry here)

I think I can get this working with (1), maybe in a follow-up PR.

mxsdev:
@sourishkrout Just pushed changes that supports arguments in the `program`. One pretty massive issue, though: annotations don't allow spaces :/ not sure what to do about that.

I also fixed error handling. You should now get the following error popup if you don't provide a `program`, and the language is not supported:

```
Unable to automatically execute cell. To execute this cell, set the "program" field in the configuration foldout!
```

And the following error popup if the user-provided `program` does not exist:

```
Unable to locate program "<program>"
```

src/client/components/configuration/annotations.ts
```diff
@@ -145,6 +145,11 @@ export class Annotations extends LitElement {
       description: "Cell's canonical name for easy referencing in the CLI.",
       docs: 'https://docs.runme.dev/configuration#cell-options',
     },
+    interpreter: {
+      description: 'Script shebang line',
+      // FIXME: update docs link
```
sourishkrout, line 150: I'll do that before releasing 👍

src/extension/runner.ts
```diff
@@ -407,6 +412,17 @@ export class GrpcRunnerProgramSession implements IRunnerProgramSession {
 
     this.session.responses.onError((error) => {
       if (error instanceof RpcError) {
+        if (error.message.includes('invalid LanguageId')) {
+          window.showErrorMessage(
+            // eslint-disable-next-line max-len
+            'Unable to automatically execute cell. To execute this cell, set the "program" field in the configuration foldout!'
```
sourishkrout, line 418: `"interpreter" field,` let's change that prior to merging please

sourishkrout (approved):
👍 LGTM, just some minor things. Feel free to change & merge.

PR #628, opened 2023-06-21
Add graphql schema

degrammer:
Add Strong typing to Grapqhl via @graphql-codegen


src/gql/createCellExecution.graphql
mxsdev, line 1: One problem here is that we have the same schema definitions duplicated across our backend and frontend. So if we change them on the backend, we'd need to remember to do it here too, otherwise we'll get runtime errors.

In our go codebase, we are using `graphql-js` to solve this problem. That package converts an introspection file to a graphql file. This is done over there like so:

```js
const { buildClientSchema, printSchema } = require("graphql");
const fs = require("fs");

const introspectionSchemaResult = JSON.parse(fs.readFileSync("introspection_query_result.json"));
const graphqlSchemaObj = buildClientSchema(introspectionSchemaResult);
const sdlString = printSchema(graphqlSchemaObj);
fs.writeFileSync("schema.graphql", sdlString);
```

I feel it would be preferable to have something like this here, too. What do you think?

src/extension/api/graphql.ts
```diff
@@ -1,27 +1,11 @@
...
 
-export interface CellExecutionRequest {
-  stdout: any
-  input: any
-  exitCode: number
-  pid: number
-  metadata: CellExecutionMetadata
-}
+export const createCellExecutionQuery = ({ data }: MutationCreateCellExecutionArgs) => {
+  const { input, exitCode, stdout, pid, metadata } = data
```
mxsdev, line 6: One concern I wonder about: What if the names on one of these properties changes? Then the name will change for the property in `data`, but we might miss changing it in the mutation call, which is not typesafe.

Do you know if it possible here to pass the `data` object directly to the mutation?

mxsdev (changes requested):

sourishkrout:
ping @degrammer. Let's make sure this PR isn't falling too far behind, please.

src/extension/api/graphql.ts
degrammer: With the new changes, this file was completely removed!

src/gql/createCellExecution.graphql
degrammer: I've removed the schema definition file, since is not necessary.

mxsdev (changes requested):
Seems like bytes are not being converted properly to strings?

https://app.staging.runme.dev/block/cljrpo2bc000zs601v17pqogk

sourishkrout:
It looks like the "Byte" scalar might be missing: https://the-guild.dev/graphql/scalars/docs/scalars/byte

mxsdev (approved):

PR #642, opened 2023-06-26
Presave check is no longer required

sourishkrout:
It dates back to when the serializer part of the parser was "experimental".

Also, tests coupled with WASM Serializer are obsolete. Serialization testing is now covered in https://github.com/stateful/runme.

PS: Relevant code makes saving fail when the markdown diff editor is active.


mxsdev (approved):
LGTM, and fixed the issue functionally as well 👍

PR #654, opened 2023-07-03
don't prompt for preexisting environment variables

mxsdev:
Ignores `export` lines if they would override an environment variable in the system store or `.env` files.

Solves #643


sourishkrout (approved):
👍 LGTM

PR #679, opened 2023-07-13
Allow referencing multiple active terminals

degrammer:


mxsdev (changes requested):
See one comment, other than that LGTM. I actually wanted to implement something like this before, but it wasn't necessitated, so this is awesome 👍

src/extension/terminal/terminalManager.ts
```diff
@@ -0,0 +1,27 @@
...
+export type ActiveTerminal = RunmeTerminal & { executionId: number }
+
+export class TerminalManager implements Disposable {
+  protected static terminals: ActiveTerminal[] = []
+
+  dispose() {
+    TerminalManager.terminals = []
+  }
+
+  static register(terminal: RunmeTerminal, executionId: number) {
```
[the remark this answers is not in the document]
degrammer: Good point, fixed!

src/extension/kernel.ts
```diff
@@ -596,4 +600,17 @@ export class Kernel implements Disposable {
   getSupportedLanguages() {
     return this.#controller.supportedLanguages
   }
+
+  registerTerminal(terminal: RunmeTerminal, executionId: number) {
+    const exists = this.activeTerminals.find((t) => t.executionId === executionId)
+    if (!exists) {
+      this.activeTerminals.push({ ...terminal, executionId })
```
mxsdev, line 607:
```suggestion
      this.activeTerminals.push(terminal)
```
One last question, is there any reason you have to destructure here? Since `executionId` hasn't changed

degrammer: but, the executionId is coming from a TaskExecution not terminal, hence we need to associate the executionId to the ActiveTerminal added to the array.

src/extension/kernel.ts
```diff
@@ -85,6 +86,7 @@ export class Kernel implements Disposable {
   protected runnerReadyListener?: Disposable
 
   protected cellManager = new NotebookCellManager(this.#controller)
+  protected activeTerminals: ActiveTerminal[] = []
```
mxsdev, line 89:
```suggestion
  protected activeTerminals = new WeakMap<string, RunmeTerminal>()
```

This might be better as a `WeakMap`, stored on current execution id. Since currently this list prevents any referenced terminals from being garbage collected, it is a potential memory leak, which WeakMap solves.

mxsdev (changes requested):

src/extension/kernel.ts
mxsdev: sorry, i realized this comment was a mistake too late. see the other comment

mxsdev (approved):

PR #680, opened 2023-07-14
Introduce terminalRows annotation

mxsdev:
Introduces `terminalRows` to select the number of rows for this cell.

In a future PR, we can consider syncing with the existing terminal, as well as edge validation cases.


sourishkrout (approved):
👍 LGTM

PatrickMayberry:
I don't think itbworked

sourishkrout:
> I don't think itbworked

Currently in the process of releasing this feature. It will work in v1.7.3, which should hit distribution in the next hour, if that's what you mean. Did I miss understand?

PR #693, opened 2023-07-17
feat:  solving dependency issues with  dependabot

duvanmonsa:
Solving conflict issue with dependabots


degrammer (approved):
LGTM

package.json
```diff
@@ -616,16 +616,16 @@
     "@graphql-codegen/cli": "4.0.1",
     "@graphql-codegen/client-preset": "^4.0.1",
     "@graphql-codegen/typescript": "4.0.1",
-    "@octokit/rest": "^19.0.8",
+    "@octokit/rest": "^20.0.1",
```
christian-bromann, line 619: This will require us to develop on Node v18 while the extension still runs on a Node v16 environment. I would suggest to keep on v19

duvanmonsa: thank you for the review yep sticking to v19

PR #700, opened 2023-07-20
persistent TLS certificate location

mxsdev:
Makes TLS location persistent. Required for certificate rotation to make a difference on the extension-side.


degrammer (approved):
LGTM

PR #713, opened 2023-07-24
Custom app config relies on base domains now

sourishkrout:
Plus additional safe-guards and tests.

Btw, merging this PR will effectively default every app user to production.


degrammer (approved):
LGTM!

PR #718, opened 2023-07-25
(chore): update wdio deps

christian-bromann:
This brings the tests back. It was likely caused by not consequently updating all WDIO dependencies.


sourishkrout (commented):
👍 Let's do it

sourishkrout (approved):

Issue #745, opened 2023-07-31
GH Action default option not respected


degrammer:
The default value configured for a GH Action is not respected, it always renders the first option.

PR #746, opened 2023-07-31
Fix gh action default value

degrammer:
Fixes https://github.com/stateful/vscode-runme/issues/745


degrammer:
@sourishkrout  tests are passing now, but that error you're getting, I'm not having it, and this has been like that since we launched the feature, can you check if you have the same issue in main?

sourishkrout:
> @sourishkrout tests are passing now, but that error you're getting, I'm not having it, and this has been like that since we launched the feature, can you check if you have the same issue in main?

it's in `main` too, weirdly enough

PR #750, opened 2023-07-31
(bugfix): fix Vercel deployment workflow

christian-bromann:
The Vercel deployment view would appear even if calling `npm i vercel`:

vercelerror

This patch fixes this error.


sourishkrout (approved):
👍 LGTM

Issue #733, opened 2023-07-28
Multiple categories aren't support within the notebook


christian-bromann:
Based on the [docs](https://docs.runme.dev/configuration#run-all-cells-by-category) a user can define multiple categories for a cell. However in VS Code the extension merges the categories together:

<img width="759" alt="Screenshot 2023-07-28 at 15 24 54" src="https://github.com/stateful/vscode-runme/assets/731337/78eb42e6-211e-4b08-bb64-f7f0c8c20847">

sourishkrout:
I really don't love that logic like this needs reimplementation per client.

Issue #591, opened 2023-06-07
Display category run button having unsaved document


duvanmonsa:
When adding a new category for a document we need to make sure we display the Run by Category button even if the document has not been saved just yet.

christian-bromann:
@duvanmonsa I am not sure how the Runme binary is suppose to pick up the category without having the user save the document first. There are sure ways to dynamically tell Runme that the user is about to add a new category but that seems to be very impractical. Can you further explain the use case why it would be valuable to the user to run by category without having the change stored to the cell?

duvanmonsa:
@christian-bromann Well I was not sure if it was possible maybe you are right and it makes no sense. it was weird to see you add a new Category but then it won't show you the option to run the cells for that category, if you feel saving the document is required that makes sense to me and we can close the ticket :)

sourishkrout:
Is it a matter of saving or reloading? If it's the former than a confirmation notification reminding the user to save to apply the change after adding a category is an imperfect but honest solution.

christian-bromann:
> Is it a matter of saving or reloading? If it's the former than a confirmation notification reminding the user to save to apply the change after adding a category is an imperfect but honest solution.

Yes, it is the former. I could see two approaches here:

- show a notification as soon as anything was changed
- show a notification when something in the cell configuration was changed and the window got closed, e.g. by clicking on `Configure` again

What do you think?

sourishkrout:
> 

Don't we know when a category was added? Only then.

christian-bromann:
> Only then.

To clarify: a user adds a category through the Notebook UI, we will have a notification popping up, reminding the user to save the document for the changes to have an impact. The popup will have a save button that on click will save the document. If the user presses `CMD`+`s` the pop-up disappears.

sourishkrout:
That sounds right. Not sure if the VS Code API lets you clear notifications (left to users exclusively) on, let's say `CMD+S`. However, even if not, that's a nice to have.

christian-bromann:
How about this:

![savewarn](https://github.com/stateful/vscode-runme/assets/731337/8ce72976-2bc0-4590-868d-d94cbe4db252)

PR #763, opened 2023-08-09
(feat): support multiple categories

### LINKED ISSUE #592: Current UI to add categories is confusing because we display and input and two buttons.
Currency logic to add categories is confusing because we display an input that displays information but doesn't work as an input confusing people when adding categories. 

Image

WE need to improve the UIX

christian-bromann:
## Current Behavior

A user can only select a single category per cell.

## Desired Behavior

Allow to set multiple categories per cell within the notebook to allow users to compose workflows from different unique cells.

## Notes

Support in CLI was implemented in https://github.com/stateful/runme/pull/354.
fixes https://github.com/stateful/vscode-runme/issues/733
fixes #592
fixes https://github.com/stateful/vscode-runme/issues/591

## Demo

__Adding existing or new categories:__
quickpick

__Warn user to save notebook before proceeding__
savewarn


christian-bromann:
@degrammer @sourishkrout any chance we can get a review on this?

sourishkrout:
> @degrammer @sourishkrout any chance we can get a review on this?

This one is for you @degrammer please.

degrammer:
There is a bug after adding a new category, the old ones disappears, this test is done against the CATEGORIES.md example 
bug

christian-bromann:
@degrammer it seems you didn't save the document. I think we don't want to save it on behalf of the user due avoid changes to the document that are irreversible which is why there is a little pop-up coming up. I recognise though that this doesn't seem to be apparent enough 🤔

src/client/components/configuration/categorySelector.ts
```diff
@@ -1,13 +1,199 @@
...
+          </a>
+        </li>
+        <li class="action-item" role="presentation" title="Remove Category Item">
+          <a
+            class="action-label codicon codicon-settings-remove"
+            role="button"
+            aria-label="Remove Category Item"
+            aria-checked=""
+            @click="${() => {
+              const event = new CustomEvent('onCategoryRemoved', {
```
degrammer, line 167: can we initialize the Custom Events at the constructor level ? and then do something like:
`this.dispatchEvent(this.onCategoryRemovedEvent)`

src/extension/kernel.ts
```diff
@@ -310,47 +307,83 @@ export class Kernel implements Disposable {
...
+            return
+          }
+
+          input.dispose()
+          postClientMessage(this.messaging, ClientMessages.onPrompt, {
+            answer: input.selectedItems.map((qp) => qp.label).join(CATEGORY_SEPARATOR),
+            uuid: message.output.uuid,
+          })
+        }),
+      )
```
degrammer, line 364: I would like to start moving out logic from the Kernel, and have their own separate file for messages logic, take a look at messages folder, this could be inside `messages/categorySelector` or similar, WDYT?

degrammer (changes requested):
It looks great, just one small comment regarding moving the kernel logic and check the bug I've found, after that's fixed, happy to approve!

christian-bromann:
@degrammer thanks for the review! I addressed the issues. Please take another look.

PR #766, opened 2023-08-11
Introduce tangle support to emit messages from client components

degrammer:


degrammer:
> LGTM 👍 . The one thing I'm wondering is if we're introducing a panel manager why not have it own all of the "tangle sync" logic which is currently dispersed, eg initializing the tangle bus, a narrow interface to emit messages etc. However, we can improve on that later.

Yeah, makes sense, I will iterate on that as a separate work.

PR #774, opened 2023-08-14
(chore): repo cleanup

christian-bromann:
This patch updates all dependencies and cleans up all dependabot PRs. This patch also includes:

- removal of `src/client/components/vite.ts` which we didn't use anymore
- downgrading `got` to v11 as everything upwards is ESM only which is not supported by VS Code


christian-bromann:
@sourishkrout there have been some changes in the cell annotation parsing to which I have no explanation for. It seems that Zods `safeParse` only returns one error of the same kind, e.g.:

```ts
const input = {
    background: 'invalid',
    interactive: 'invalid',
    closeTerminalOnSuccess: 'invalid'
}
const parseResult = CellAnnotationsSchema.safeParse(input) as SafeParseError<any>
expect(parseResult.success).toBeFalsy()
const { fieldErrors } = parseResult.error.flatten()
expect(fieldErrors).toEqual({ background: ['expected a boolean value'] })
```

and

```ts
const input = {
    // background: 'invalid', <-- removing background
    interactive: 'invalid',
    closeTerminalOnSuccess: 'invalid'
}
const parseResult = CellAnnotationsSchema.safeParse(input) as SafeParseError<any>
expect(parseResult.success).toBeFalsy()
const { fieldErrors } = parseResult.error.flatten()
expect(fieldErrors).toEqual({ interactive: ['expected a boolean value'] }) // now getting this error
```

I am not sure if this was caused by one of the dependency updates or due to other changes. I also don't know if it has an impact on the general functionality of the extension.

christian-bromann:
@sourishkrout after further investigations I pushed more changes that would address the following problems:

- I removed old code we don't use anymore to reduce having to deal with more dependencies (e.g. removed the alternative Vercel implementation as well as the Vite cell renderer I used for rendering of HTML)
- clear out `node_modules` folder before running e2e tests to ensure we capture bugs where transient dependencies are imported through dynamic `require` calls or where any file is read via `fs`
- since this isn't an obvious procedure I documented our testing stack in `Contributing.md` and made sure we execute and reconcile the dependency folder after the test finishes

I think this is a great example showing how useful Runme can be since putting all this un-documented in a `package.json` could have caused confusion later on.

Please take another look!

sourishkrout:
> I am not sure if this was caused by one of the dependency updates or due to other changes. I also don't know if it has an impact on the general functionality of the extension.

Don't think it'll cause downstream problems. Tested it. The only way to can multiple invalid values in the schema is by directly editing the annotations. However, if we can validate them the UI will use default values with a warning.

sourishkrout (approved):
LGTM 👌 good to see the e2e tests will catch bundling errors now.

christian-bromann:
Let's also validate that all telemetry is send correctly as we had a major update for that package too.

sourishkrout:
> Let's also validate that all telemetry is send correctly as we had a major update for that package too.

Coming thru in the latest pre-release (edge).

PR #775, opened 2023-08-14
(bugfix): recognise dot entries in `.gitignore`

christian-bromann:
This patch ensures that any gitignore entries with a starting dot are recognised.

__Behavior Before:__
Screenshot 2023-08-14 at 14 29 19

__Behavior After:__
Screenshot 2023-08-14 at 14 28 13


degrammer (approved):
LGTM

PR #792, opened 2023-08-22
(bugfix): don't prompt for env if set to false

### LINKED ISSUE #778: promptEnv=false is asking me to set env variables
Hello guys.

I tried to run this simple example in VS Code and it is still asking me to set Environment Variables even when I use `promptEnv=false`:

<pre>
```sh { promptEnv=false }
export AWS_DEFULT_REGION="eu-central-1"
echo "${AWS_DEFULT_REGION}"
```
</pre>

image

I would like to set the environment variables without user interaction (because if there are many of them - then this is a real issue)

Thank you for looking at it...

christian-bromann:
fixes #778

### DIFF
--- a/examples/test.md
+++ b/examples/test.md
@@ -12,6 +12,10 @@ This markdown file contains some custom examples to test the execution within a 
 echo "Hello World!"
 ```
 
+```sh { promptEnv=false }
+export FOO="don't prompt me"
+```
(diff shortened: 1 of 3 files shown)


degrammer (approved):
Tested  and LGTM 🚀

PR #794, opened 2023-08-22
Align @types/node with what is supported in Electron

christian-bromann:
After having internal discussions about this , let's stick with the version that is actually run within Electron.

Closes #793


sourishkrout (approved):
👍 LGTM

PR #796, opened 2023-08-22
(test): retries test files at least twice

christian-bromann:
While just retrying isn't the ideal solution, if it unblocks us for now it is worth moving forward with that as I don't see any consistent flakes yet.


christian-bromann:
This seems to be much better now. I will go ahead and unblock the build.

PR #798, opened 2023-08-22
Consolidate Terminal Styles

### LINKED ISSUE #611: Runme terminal settings
Runme Terminal, should allow more customization, so far we support Font family and Font size, and there are a bunch of terminal settings that can be provided like:

- Scrollback: Amount of retained rows when lines are scrolled beyond the initial viewport
- Fast scroll sensitivity: The scroll speed multiplier used for fast scrolling
-  Cursor style
- Cursor width
Review what other settings can be supported

christian-bromann:
Fixes #611

### DIFF
--- a/package.json
+++ b/package.json
@@ -453,6 +453,34 @@
             "default": null,
             "markdownDescription": "Font family of Runme notebook terminal"
           },
+          "runme.terminal.cursorStyle": {
+            "type": "string",
+            "enum": [
+              "block",
+              "bar",
+              "underline"
+            ],
+            "default": "bar",
+            "markdownDescription": "The style of the Runme notebook terminal cursor."
+          },
+          "runme.terminal.cursorBlink": {
+            "type": "boolean",
+            "default": true,
+            "markdownDescription": "Defines whether the Runme notebook terminal cursor blinks."
+          },
+          "runme.terminal.cursorWidth": {
+            "type": "number",
+            "markdownDescription": "The width of the cursor in CSS pixels when `cursorStyle` is set to 'bar'."
+          },
+          "runme.terminal.smoothScrollDuration": {
+            "type": "number",
+            "default": 0,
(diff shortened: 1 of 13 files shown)


christian-bromann:
This is now ready for review.

package.json
```diff
@@ -453,6 +453,34 @@
...
+          "runme.terminal.cursorWidth": {
+            "type": "number",
+            "markdownDescription": "The width of the cursor in CSS pixels when `cursorStyle` is set to 'bar'."
+          },
+          "runme.terminal.smoothScrollDuration": {
+            "type": "number",
+            "default": 0,
+            "markdownDescription": "The duration to smoothly scroll between the origin and the target in milliseconds. Set to 0 to disable smooth scrolling and scroll instantly."
+          },
+          "runme.terminal.scrollback": {
```
degrammer, line 480: I think is better to have a default value of 100, see integrated terminal settings for VS Code

degrammer (approved):
LGTM, I've added some suggestion to the default scrollback setting, also, for a next iteration, I imagine the cursor width setting with something like this:
Screenshot 2023-09-13 at 9 59 51 AM

package.json
christian-bromann: Currently the implemented behavior mimics what we have in `main` which is not setting the value at all and have it default to the [default value](https://github.com/xtermjs/xterm.js/blob/bdb25c2dbee65af8afa36dca816af51fe5ab1d5d/src/common/services/OptionsService.ts#L33). Are you suggesting to explicitly set the default value to `100`?

christian-bromann:
> I imagine the cursor width setting with something like this

As mentioned above: the current behavior will not set the value when undefined, so we rely on the xTerm default value.

christian-bromann:
> I imagine the cursor width setting with something like this

As mentioned above: the current behavior will not set the value when undefined, so we rely on the xTerm default value.

PR #812, opened 2023-09-06
feat: get blocks from runme binary

duvanmonsa:
Adding changes to the launcher to use runner

### DIFF
--- a/src/extension/provider/launcher.ts
+++ b/src/extension/provider/launcher.ts
@@ -1,21 +1,22 @@
-import { join, basename, dirname, resolve } from 'node:path'
+import { basename, dirname, join, resolve } from 'node:path'
 
 import {
-  TreeItem,
-  EventEmitter,
-  workspace,
-  TreeItemCollapsibleState,
   Command,
-  TreeDataProvider,
+  Disposable,
   Event,
+  EventEmitter,
+  FileType,
+  TreeDataProvider,
+  TreeItem,
+  TreeItemCollapsibleState,
   Uri,
   commands,
-  FileType,
-  Disposable,
+  workspace,
 } from 'vscode'
 
 import { Kernel } from '../kernel'
-import { mapGitIgnoreToGlobFolders, getPathType } from '../utils'
+import { IRunner } from '../runner'
+import { getPathType, mapGitIgnoreToGlobFolders } from '../utils'
 
 interface IRunmeFileProps {
   tooltip: string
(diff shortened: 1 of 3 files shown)


src/extension/runner.ts
```diff
@@ -244,6 +248,34 @@ export class GrpcRunner implements IRunner {
...
+          return blocks
+        })
+        .catch((e) => {
+          throw e
+        })
+    } catch (err: any) {
+      console.error(err)
+      this.close()
+      throw err
+    }
```
christian-bromann, line 276:
```suggestion
      const { response } = await client.getBlocks(request)
      return response.blocks || []
    } catch (err: any) {
      console.error(err)
      this.close()
      throw err
    }
```

src/extension/provider/launcher.ts
```diff
@@ -58,10 +59,16 @@ export class RunmeFile extends TreeItem {
 export class RunmeLauncherProvider implements TreeDataProvider<RunmeFile>, Disposable {
   #disposables: Disposable[] = []
 
+  protected runner?: IRunner
+
```
christian-bromann, line 63:
```suggestion
```

src/extension/provider/launcher.ts
```diff
@@ -58,10 +59,16 @@ export class RunmeFile extends TreeItem {
...
   private filesTree: Map<string, TreeFile> = new Map()
   private defaultItemState: TreeItemCollapsibleState = TreeItemCollapsibleState.Collapsed
 
-  constructor(private workspaceRoot?: string | undefined) {
+  constructor(
+    runner: IRunner | undefined,
```
christian-bromann, line 68: I think there is never a case where we don't provide the runner, right?
```suggestion
    protected runner: IRunner,
```

src/extension/provider/launcher.ts
```diff
@@ -156,6 +163,17 @@ export class RunmeLauncherProvider implements TreeDataProvider<RunmeFile>, Dispo
     this.refresh()
   }
 
+  private async loadBlocks() {
+    try {
+      const blocks = await this.runner?.getBlocks()
```
christian-bromann, line 168:
```suggestion
      const blocks = await this.runner.getBlocks()
```

src/extension/provider/launcher.ts
duvanmonsa: I was trying to follow this example here.
https://github.com/stateful/vscode-runme/blob/b242c5e71a39d8cd4063c2f7a20ce2eb9fa9deca/src/extension/extension.ts#L62

sourishkrout:
Closing since we now have https://github.com/stateful/runme/pull/438 and subsequent work on the project service.

Issue #776, opened 2023-08-15
Fix save output to rely on NotebookCellOutput instead of terminal state


sourishkrout:
Scenario to reproduce is:

In a cell execute:

```
echo "Hello World to Slack!"
exit 111
```

Then change the `exit 111` to `exit 0` and rerun.

On Save, it will continue to report 111 instead of a failed execution (zero code not populated because it never executed). We default to 0 in this case which is kind of the worst because it signals success when it wasn't.

Proposed fix:

1. Exclusively use NotebookCellOutput* to retain execution information ahead of saving.
2. When saving let's not default to a exit code if none is present due to runnerExitStatus.
3. Push runnerExitStatus to server ideally grabbed from NotebookCellOutput*.

![Image](https://github.com/stateful/vscode-runme/assets/250527/7507c0da-7f7d-4f5d-970f-61e4517d1a85)

sourishkrout:
Done https://github.com/stateful/vscode-runme/pull/829.

PR #815, opened 2023-09-07
feat: errorCode from execution output

duvanmonsa:
Passing errorCode from execution runner before saving the file.

### DIFF
--- a/src/client/components/terminal/index.ts
+++ b/src/client/components/terminal/index.ts
@@ -305,6 +305,7 @@ export class TerminalView extends LitElement {
   protected windowSize: IWindowSize
 
   protected rows: number = 10
+  protected errorCode?: any
 
   @property({ type: String })
   uuid?: string
@@ -398,6 +399,13 @@ export class TerminalView extends LitElement {
         switch (e.type) {
           case ClientMessages.activeThemeChanged:
             this.#updateTerminalTheme()
+            break
+          case ClientMessages.terminalErrorCode:
+            const { 'runme.dev/uuid': uuid, code } = e.output
+            if (uuid !== this.uuid) {
+              return
+            }
+            this.errorCode = code
             break
           case ClientMessages.terminalStdout:
           case ClientMessages.terminalStderr:
@@ -703,6 +711,7 @@ export class TerminalView extends LitElement {
       await postClientMessage(ctx, ClientMessages.cloudApiRequest, {
         data: {
           stdout: contentWithAnsi,
+          errorCode: this.errorCode,
         },
         uuid: this.uuid!,
         method: APIMethod.CreateCellExecution,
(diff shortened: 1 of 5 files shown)


duvanmonsa:
https://github.com/stateful/vscode-runme/issues/776

src/client/components/terminal/index.ts
```diff
@@ -305,6 +305,7 @@ export class TerminalView extends LitElement {
   protected windowSize: IWindowSize
 
   protected rows: number = 10
+  protected errorCode?: any
```
christian-bromann, line 308: We know the type of this one right?
```suggestion
  protected errorCode?: number
```

src/client/components/terminal/index.ts
```diff
@@ -305,6 +305,7 @@ export class TerminalView extends LitElement {
   protected windowSize: IWindowSize
 
   protected rows: number = 10
+  protected errorCode?: number | void
```
christian-bromann, line 308: This doesn't work?
```suggestion
  protected errorCode?: number
```

duvanmonsa: @christian-bromann nop I tried it but it was not working where we get the errorCode returns that type I can for the type if you want. 
image

christian-bromann: ok, I think it is fine then to keep as is.

degrammer (approved):
LGTM, despite this approach works, let's check with @sourishkrout  about this.

sourishkrout:
> Passing errorCode from execution runner before saving the file.

What GitHub issue is this solving? Just for context.

admc:
[I think it's this](https://github.com/stateful/cloud/issues/117)

duvanmonsa:
@sourishkrout my bad yes this one https://github.com/stateful/vscode-runme/issues/776

duvanmonsa:
> [I think it's this](https://github.com/stateful/cloud/issues/117)

that one is different @admc that's on the FE for cloud.

sourishkrout:
> @sourishkrout my bad this one #776

This is definitely just a bandaid so depending on when we can expect a full resolution of #776 (no longer relying on terminal state) we can decided whether or not to merge. What prompted this intermediate fix? Unless there's a good reason I feel like we should fix this once and for all instead of spending time on surface-level bandaids.

sourishkrout:
> > [I think it's this](https://github.com/stateful/cloud/issues/117)
> 
> that one is different @admc that's on the FE for cloud.

It's likely both.

duvanmonsa:
Then maybe we don't know how to solve it correctly, Right now we are getting the errorCode directly from the response of the execution and it was working as the expected behavior.  I would need some help to see how we can solve this in the proper way.

degrammer:
Closing in favor of https://github.com/stateful/vscode-runme/pull/829

PR #829, opened 2023-09-13
Fix Terminal state

degrammer:
Each time a new cell execution is triggered, if there is any existing executed integrated terminal, it should be removed and disposed, otherwise, it can create inconsistencies when referencing an active terminal by runme id.

We only need one single reference to the last active terminal per cell.


sourishkrout (approved):
Let's merge this since it's a pragmatic fix for #776. Once merged we can just close https://github.com/stateful/vscode-runme/pull/815 without merging, no?

PR #797, opened 2023-09-13
Pressing `o` in the cell configuration panel collapses the panel

christian-bromann:
It seems that VS Code has a keybinding for closing cells when pressing the `o` key:

bugdemo

This is not an ideal user experience when editing cell configurations.


christian-bromann:
It turns out that this behavior originates from the [Jupyter Keymap](https://github.com/microsoft/vscode-jupyter-keymap/blob/main/package.json#L32-L126) extension and the only way to suppress this is by overwriting the keybindings. I raised [an issue](https://github.com/microsoft/vscode-jupyter-keymap/issues/32) asking whether they could very the kernel before executing the command to allow us to remove these overwrites.

sourishkrout:
I'm getting a warning that the `noop` command does not exist, @christian-bromann?

image

sourishkrout (changes requested):
Mergable once the warning's removed (see previous comments).

sourishkrout:
> @sourishkrout I don't see that warning on my end, can you please check again? this seems GTG

Confirmed 👍

PR #832, opened 2023-09-15
Add GitHub secret for tests

degrammer:
I'm not 100% If this is the proper place to add the secret to dependabot.yml, let's test. We need to add it to the dependabot secrets @christian-bromann

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,23 +1,25 @@
 version: 2
 updates:
-- package-ecosystem: npm
-  directory: "/"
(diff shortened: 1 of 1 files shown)


christian-bromann:
@degrammer can you reference the docs that describes to add secrets this way?

degrammer:
> @degrammer can you reference the docs that describes to add secrets this way?

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/configuring-access-to-private-registries-for-dependabot

PR #841, opened 2023-09-19
Fix XTerm deps

degrammer:
Dependabot fix for https://github.com/stateful/vscode-runme/pull/826

It needs further testing


sourishkrout (approved):
LGTM 👍 . Let's merge. We can "break it in" for a few days in pre-release.

Issue #830, opened 2023-09-13
Document level promptEnv setting


iamprakkie:
I have use case where steps to deploy solution are README. These steps involve exporting env variables with outcomes of command. For example, 
```bash
export COA_AMG_API_KEY=$(aws grafana create-workspace-api-key --profile monitoring-account --region ${COA_MON_REGION} \
    --key-name "grafana-operator-key" \
    --key-role "ADMIN" \
    --seconds-to-live 432000 \
    --workspace-id $COA_AMG_WORKSPACE_ID \
    --query key \
    --output text)
```

I need to disable `promptEnv` for this cell, this command runs fine. However, I have several cells like this and need to disable for every cell. Option to change `promptEnv` at document level will be helpful for this scenario.

sourishkrout:
Thanks for raising the issue @iamprakkie! The feature request makes perfect sense.

We will try to get this into Runme in the next few weeks.

edeediong:
Hi @iamprakkie  thank you for bringing this up. I can confirm that the [feature](https://docs.runme.dev/configuration) you mentioned is still present in our latest pre-release version (v1.11.1698245570). If you'd like to try it out, you can switch to the [pre-release version](https://docs.runme.dev/install) using the link provided. Please let us know if you have any feedback or questions. Thank you!

PR #843, opened 2023-09-21
Enable / Disable toggle envPrompt for Notebook

degrammer:
This PR introduces two new commands that allows to enable /disable skip the promptEnv setting for all the Notebook cells, this will prevent going cell by cell to disable the setting when is needed in certain operations.

Screenshot 2023-09-20 at 10 35 26 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -184,8 +184,13 @@
         "category": "Runme"
       },
       {
-        "command": "runme.skipPromptEnv",
-        "title": "Skip prompt environment variables",
+        "command": "runme.enableSkipPromptEnv",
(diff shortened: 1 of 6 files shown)


sourishkrout (changes requested):
@degrammer is this solving https://github.com/stateful/vscode-runme/issues/830?

I think you misunderstood the requirement here. The idea is to have a top-level `promptEnv: false` property in the frontmatter that will disable this. Not VS Code commands.

sourishkrout:
Moving this to draft until comments above are addressed @degrammer.

PR #867, opened 2023-10-02
Fix cursor width for terminal

degrammer:
The default cursor width should be 1 instead of 0


sourishkrout (approved):
Fix confirmed 👍 . LGTM.

PR #871, opened 2023-10-04
Fix cell execution

degrammer:
Cell execution is more reliable now, this approach seems to fix the kernel selection issue too.
Also, global state update was missing corresponding async, it was causing some race conditions.

### DIFF
--- a/tests/e2e/specs/categories.e2e.ts
+++ b/tests/e2e/specs/categories.e2e.ts
@@ -25,6 +25,7 @@ async function removeAllNotifications () {
   await Promise.all(notifications.map((notification) => notification.dismiss()))
 }
 
+// TODO: Fix this test
 describe.skip('Runme Categories Tests', async () => {
   const notebook = new RunmeNotebook()
   const webview = new Webview()
--- a/tests/e2e/wdio.conf.ts
+++ b/tests/e2e/wdio.conf.ts
@@ -94,7 +94,7 @@ export const config: Options.Testrunner = {
     //
     capabilities: [{
         browserName: 'vscode',
-        browserVersion: '1.82.3',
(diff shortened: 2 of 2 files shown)


src/extension/handler/uri.ts
```diff
@@ -98,7 +98,7 @@ export class RunmeUriHandler implements UriHandler {
           const isProjectOpened =
             workspace.workspaceFolders?.length &&
             workspace.workspaceFolders.some((w) => w.uri.path === projectPath.path)
-          this.#context.globalState.update(EXECUTION_CELL_STORAGE_KEY, cell)
+          await this.#context.globalState.update(EXECUTION_CELL_STORAGE_KEY, cell)
           if (!isProjectOpened) {
             await writeDemoBootstrapFile(projectPath, fileToOpen, cell)
             await commands.executeCommand('vscode.openFolder', projectPath, {
```
sourishkrout, line 104: Can we make `forceNewWindow` a setting please? I think we want it to be default `true` as opposed to `false` (which makes dev testing easier).

sourishkrout (changes requested):
Minor request but otherwise 👍 LGTM.

src/extension/extension.ts
```diff
@@ -205,7 +205,7 @@ export class RunmeExtension {
         const cell = context.globalState.get<number>(EXECUTION_CELL_STORAGE_KEY)
```
sourishkrout, line 205: It might be wise to not just add the cell but also a timestamp, potentially discard the operation if the entry is e.g. >5min old. Otherwise this could lead to weird side effects.

sourishkrout (approved):
Works. Good to merge 👍 . Let's put the remaining changes in a new PR.

PR #874, opened 2023-10-06
Introduce file watcher to already opened folder

degrammer:


src/extension/handler/utils.ts
```diff
@@ -225,3 +225,37 @@ export async function cleanExecutionDemo(context: ExtensionContext) {
   await context.globalState.update(EXECUTION_CELL_STORAGE_KEY, undefined)
   await context.globalState.update(EXECUTION_CELL_CREATION_DATE_STORAGE_KEY, undefined)
 }
+
+export function createDemoFileRunnerWatcher(context: ExtensionContext, kernel: Kernel) {
```
sourishkrout, line 229: I'm not sure if the functional pattern expanding utils is the best paradigm here. It feels like sub-classing per command might be a good approach. However, let's refactor/rewrite if we decide to move `demo` to something more stable.

PR #905, opened 2023-10-17
Fix the toolbar icons in dark mode

admc:
I guess the think magic for the toolbar icons only works in insiders at the moment


sourishkrout:
> I guess the think magic for the toolbar icons only works in insiders at the moment

Magic @admc?

admc:
in insiders the stroke and fill are being automatically replaced by the theme css variable, but not in the release version, you need to use white and black for dark and light. So let's do that for now

sourishkrout (approved):
👍 LGTM

PR #919, opened 2023-10-23
Add missing time for cell execution

degrammer:
This fix allows cell execution inspection time via **cell.executionSummary.timing.endTime** and **cell.executionSummary.timing.startTime**, without an end execution, we were missing that data.


sourishkrout (approved):
LGTM 👍 . Please be sure to close the loop with the user once merged. Perhaps after making a pre-release, @degrammer.

PR #922, opened 2023-10-23
Rollback vscode-uri version

degrammer:
The last version of vscode-uri is breaking the e2e tests, I think this is related to the way [wdio-vscode-service](https://github.com/webdriverio-community/wdio-vscode-service) handles the [dependency](https://github.com/webdriverio-community/wdio-vscode-service/blob/main/src/server/utils.ts#L6)

They're switching to package.json exports https://github.com/microsoft/vscode-uri/blob/53e4ca6263f2e4ddc35f5360c62bc1b1d30f27dd/package.json#L8-L12

Screenshot 2023-10-23 at 6 10 06 PM

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -21,3 +21,5 @@ updates:
       # following dependencies dropped support for Node.js v16
       # which is the current runtime for VS Code extensions
       - dependency-name: "@octokit/rest"
+      # it breaks wdio-vscode-service
+      - dependency-name: "vscode-uri"


christian-bromann:
Very interesting! Unfortunately I currently don't have the capacities to contribute to the VS Code service but any contributions are welcome 😉

sourishkrout (changes requested):
We might have to include the rolled back dep into https://github.com/stateful/vscode-runme/blob/main/.github/dependabot.yml to prevent it from getting bumped again. Also, let's create a follow-up ticket to circle back on the issue later @degrammer.

sourishkrout (approved):
👍 @degrammer

PR #935, opened 2023-11-02
Introducing terminal output auto-save

degrammer:
Introducing terminal output auto-save!

- Automatically record the terminal output
- Added as extension setting
- Currently the option is enabled / disabled globally for all the notebooks
- Update Graphql schema to latest

auto-save


sourishkrout (approved):
Works! 👍 LGTM.

PR #938, opened 2023-11-03
Migrates all logic relying on ephemeral runme.dev/uuid to persistent id

pastuxso:
Resolve: https://github.com/stateful/runme/issues/408


sourishkrout:
Let's hold off merging until after we ship release v2 (https://github.com/stateful/runme/pull/412). While this PR is short-circuiting the `uuid` <> `ulid` migration, I'd like to take some time to do a complete rename/consolidation of the uuid stuff.

From a user's perspective, there is no difference in whether a transient or persistent ID is used for cell identity until features require them to be constant. Let's embrace this for a short transition phase @pastuxso & @joseeantonior.

PR #962, opened 2023-11-13
transfer examples to right place

edeediong:


sourishkrout (changes requested):
how about `examples/hardening/server.md` @edeediong?

sourishkrout (approved):
👍 LGTM

examples/shebang.md
```diff
@@ -14,16 +14,16 @@ print("Hello Pythonista 🐍")
 ```rb
 # Open up the String class to add a new method
 class String
-    # Define a new method to reverse the words in a string
```
sourishkrout: @edeediong this didn't seem right

PR #980, opened 2023-11-25
Reactive stable and remove version pinning

sourishkrout:
Try latest stable removing pinned version. Is this still an issue?


pastuxso:
@sourishkrout unfortunately is still failing, seems the function **openTerminalView()** is not compatible with the latest version of vscode on Linux.

cc: @degrammer :point_up:

PR #998, opened 2023-12-08
Project service integration

sourishkrout:
Replace old/unfinished task provider integration with GRPC project service.

This is an intermediate step to make tasks more integral between "project mode" and VS Code's task system. What's missing but slated for follow up PRs:
- [ ] Reconcile task `newRunmeProjectTask` vs `newRunmeTask` should be able to have a single impl.
- [ ] Remodel the `runme` TaskDefinition to closer match what cells are now.
- [ ] Integrate tasks with "Runme Notebooks" tree view.
- [ ] Refresh cached tasks when fs changes occur and implement `resolveTask` to trigger a refresh.

To test this: Open the command palette and search for "Tasks: Run Task", select "runme" from the list. The list should match what you'd see (only named commands) if you run `runme ls` in the root of the repo. Select an entry to execute.

### DIFF
--- a/__mocks__/vscode.ts
+++ b/__mocks__/vscode.ts
@@ -29,6 +29,7 @@ export class Uri extends URI {
 }
 
 export const workspace = {
+  asRelativePath: vi.fn(),
   getConfiguration: vi.fn().mockReturnValue(new Map()),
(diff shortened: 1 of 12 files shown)


pastuxso (approved):
LGTM

PR #1008, opened 2023-12-11
Improve serialization behavior/timing to avoid accidental data-loss

sourishkrout:
The VS Code API provides limited control over serialization timing; however, this PR improves two things that were possible:

1. The session outputs file will be written on saving and serializing. The latter is new. This allows a near real-time (vscode debounces/throttles serialization, which introduces a slight lag) tracking of state side-by-side. This still feels more natural.
2. The notebook document is marked "dirty" (needs to be saved before closing/exiting) after every successful execution. This makes it explicit always to save when a "mutation" happens.

The changes feel natural and fix the problems described [here](https://github.com/stateful/runme/issues/400#issuecomment-1849030175) where serializing outputs to the session file do not match (out of sync) the latest notebook state on closing/exiting.

To test this: without this PR the last run of a cell won't be written to the session file after the execution completes until the user explicitly triggers save (CTRL+S / CMD+S). With this PR, the document will be marked dirty as part of completing a cell run requiring a save/discard by the user, making the save-requirement explicit.

### DIFF
--- a/src/extension/cell.ts
+++ b/src/extension/cell.ts
@@ -34,7 +34,10 @@ const NOTEBOOK_SELECTION_COMMAND = '_notebook.selectKernel'
 export class NotebookCellManager {
   #data = new WeakMap<NotebookCell, NotebookCellOutputManager>()
 
-  constructor(protected controller: NotebookController) {}
+  constructor(
+    protected controller: NotebookController,
+    protected readonly outputPersistence: boolean,
+  ) {}
 
   registerCell(cell: NotebookCell): NotebookCellOutputManager {
     const existing = this.#data.get(cell)
@@ -42,7 +45,7 @@ export class NotebookCellManager {
       return existing
     }
 
-    const outputs = new NotebookCellOutputManager(cell, this.controller)
+    const outputs = new NotebookCellOutputManager(cell, this.controller, this.outputPersistence)
     this.#data.set(cell, outputs)
 
     return outputs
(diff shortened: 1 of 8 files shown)


src/extension/cell.ts
```diff
@@ -319,10 +324,9 @@ export class NotebookCellOutputManager {
       }
 
       this.onFinish = new Promise<void>(async (resolve) => {
-        // todo: see bug note inside refreshTerminal
```
pastuxso, line 322: Could you please educate me on why we are adding this block of code again?

pastuxso (approved):
LGTM

src/extension/cell.ts
sourishkrout: Of course. The code was initially written to make outputs serialization available to refresh a secondary output item (`stdout`) with the state of the notebook terminal's stdout/stderr to be sent to the kernel. However, the strategy broke down when I ran into an issue with the VS Code API `replaceOutput`, which appears to be buggy still and, instead of replacing, does concatenate (wind up with 3 instead of 2 output items). The serializer now backfills the stdout at serialization time instead (which ain't great but works).

At the time, I decided to comment out the code because I knew we could reuse it for the exact problem it's now being used to solve, which is marking the document as "dirty" to signal to the editor it needs saving. I knowingly left this problem unsolved at the time to get an early version shipped.

You'll find another comment block inside the method now explaining the strategy change.

PR #1028, opened 2023-12-29
sonarcloud fine-tuning

pastuxso:


.github/workflows/test.yml
```diff
@@ -1,5 +1,9 @@
 name: Tests
-on: [push, pull_request]
+on:
+  pull_request:
```
pastuxso, line 3: I moved this one in two separate lines to avoid twice Github Action runs,

vitest.conf.ts
```diff
@@ -9,11 +9,11 @@ export default defineConfig({
...
     ],
     coverage: {
-      reporter: ['html', 'lcov'],
+      reporter: ['lcov'],
       enabled: true,
-      exclude: ['**/build/**', '**/__fixtures__/**', '**/*.test.ts'],
+      exclude: ['**/build/**', '**/__fixtures__/**', '**/*.test.ts', '**/__mocks__/**'],
       statements: 43,
-      branches: 85,
+      branches: 80,
```
pastuxso, line 16: Decreased to 80 to pass the current coverage (84%)

vitest.conf.ts
```diff
@@ -9,11 +9,11 @@ export default defineConfig({
       '**/node_modules/**', 'examples/**'
     ],
     coverage: {
-      reporter: ['html', 'lcov'],
+      reporter: ['lcov'],
```
pastuxso, line 12: Switched to **lcov**, the output is the same and also produces an HTML one.

vitest.conf.ts
sourishkrout: How about 84%, @pastuxso, giving up 5% is a big drop

sourishkrout (changes requested):
See comment, otherwise good 👍.

vitest.conf.ts
pastuxso: I'm going to change it to 84% :wink:

PR #1032, opened 2024-01-05
Tasks relative display paths

sourishkrout:
The project service will return absolute paths exclusively. However, in the UI we want them to be displayed relative to the workspace root(s) wherever possible. This PR introduces the baseline logic to do exactly that.


sourishkrout:
@pastuxso please review. Going to skip Sonar on this one since this code will get another round of refactoring plus tests (shipping this incrementally).

pastuxso (approved):
LGTM

PR #1075, opened 2024-01-23
Fix docs links

degrammer:
Updating some outdated / broken links from cell configuration docs

### DIFF
--- a/src/client/components/configuration/annotations.ts
+++ b/src/client/components/configuration/annotations.ts
@@ -140,7 +140,7 @@ export class Annotations extends LitElement {
     },
     mimeType: {
       description: "Cell's output content MIME type.",
-      docs: 'https://docs.runme.dev/configuration/reference#supported-mime-types',
+      docs: 'https://docs.runme.dev/configuration#supported-mime-types',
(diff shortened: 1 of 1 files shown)


sourishkrout (changes requested):
I'm wondering if we should have the extension point at some stable redirection links for easy adjustment plus the ability to retroactively fix older versions too if we break any of the destination links?

E.g. like this https://github.com/stateful/runme.dev/blob/main/next.config.js#L41-L52 but maybe with a `docs/` base @degrammer? Btw, it's intentional that the redirects are on `runme.dev` not `docs.runme.dev` because the latter uses docusaurus and does not provide HTTP-layer redirects.

sourishkrout:
> What if we use a link shortener service like https://dub.co/ ? we can associate a custom domain for that

We don't want another thing to maintain and document. We are already using the next/vercel redirects.

sourishkrout:
@degrammer please consolidate this PR with https://github.com/stateful/vscode-runme/pull/1089. It shouldn't take a lot of time and we want to include this into the next release.

PR #1077, opened 2024-01-25
Rewire autosave

degrammer:
Revamp auto-save experience to deeply integrate with session output persistence.

✏️ 
- It removes runme.experiments.outputPersistence flag since now everything is controlled via auto-save, it works now alongside runme.app.sessionOutputs
- It introduces a command to clean up login dismiss alert for un-authenticated users

### DIFF
--- a/package.json
+++ b/package.json
@@ -929,6 +929,7 @@
     "expect-webdriverio": "^4.2.7",
     "fork-ts-checker-webpack-plugin": "^9.0.0",
     "husky": "^8.0.3",
+    "msw": "^2.1.5",
     "npm-run-all": "^4.1.5",
     "prettier": "^3.2.4",
     "runme": "^3.1.0",
--- a/__mocks__/vscode.ts
+++ b/__mocks__/vscode.ts
@@ -24,6 +24,12 @@ export class Uri extends URI {
   static parse = vi.fn(super.parse)
 
   static joinPath = vi.fn((uri: Uri, ...paths: string[]) => {
+    /**
+     * Allow testing against http endpoints
+     */
+    if (uri.authority?.includes('.runme.dev')) {
+      return  `https://testing.runme.dev${path.join(...paths)}`
+    }
     return super.file(path.join(uri.fsPath, ...paths))
   })
 }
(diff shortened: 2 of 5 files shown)


src/extension/extension.ts
```diff
@@ -176,7 +177,6 @@ export class RunmeExtension {
       ...this.registerPanels(kernel, context),
       ...surveys,
       workspace.registerNotebookSerializer(Kernel.type, serializer, {
-        transientOutputs,
```
sourishkrout, line 179: why remove this? when outputs are disabled they should be transient again

package.json
```diff
@@ -673,6 +694,12 @@
             ],
             "default": "no",
             "markdownDescription": "Automatically save cell execution output"
+          },
+          "runme.app.outputPersistence": {
```
sourishkrout, line 698: rename to `runme.app.sessionOutputs`?

src/extension/serializer.ts
```diff
@@ -531,7 +545,7 @@ export class GrpcSerializer extends SerializerBase {
     const lid = GrpcSerializer.getDocumentLifecycleId(data.metadata)
     const serialRequest = <SerializeRequest>{ notebook }
 
-    const output = this.cacheNotebookOutputs(notebook, lid)
+    const output = await this.cacheNotebookOutputs(notebook, lid)
```
sourishkrout, line 548: @degrammer why `await` here? this will cause 548 and 549 to no longer run in parallel which is being awaited on 552.

src/extension/utils.ts
```diff
@@ -595,3 +603,46 @@ export function asWorkspaceRelativePath(documentPath: string): {
...
+  context: ExtensionContext,
+): Promise<AuthenticationSession | undefined> {
+  let session = await getAuthSession(false)
+  const displayLoginPrompt = await context.globalState.get(SAVE_CELL_LOGIN_CONSENT_STORAGE_KEY)
+  if (!session && displayLoginPrompt !== false) {
+    const option = await window.showInformationMessage(
+      `Securely store your cell output on the Runme Cloud.
+      Authenticating with GitHub is required, do you want to proceed?`,
+      'Yes',
+      'Dismiss',
```
sourishkrout, line 623: Let's rename `Dismiss` to `No`.

sourishkrout (changes requested):
@degrammer please address remaining comments

sourishkrout (approved):
👍 LGTM. The test coverage is inaccurate since it doesn't limit the diff to only what has changed.

PR #1078, opened 2024-01-25
Introduce infra + cloud terminology

sourishkrout:
This is mainly for VS Code marketplace visibility.

PS: The keyword "runme" when the extension is called "runme" seems redundant.


admc:
Heck yea!

admc (approved):
👍

PR #1088, opened 2024-01-29
Remove await

degrammer:


sourishkrout (approved):
👍 LGTM. Probably harmless since the TS compiler likely just removes the redundancy.

PR #1090, opened 2024-02-01
Use the gif for the new video

admc:


admc:
@sourishkrout think this makes sense?

sourishkrout (approved):
LGTM 👍

PR #1092, opened 2024-02-03
Try stable and see if chromedriver is available now

sourishkrout:
Unfortunately there's always a lag after a new VS Code major release comes out.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -84,7 +84,7 @@
         "unzipper": "^0.10.14",
         "vitest": "^1.2.2",
         "vscode-uri": "^3.0.7",
-        "wdio-vscode-service": "^5.2.3",
+        "wdio-vscode-service": "^6.0.0",
         "webdriverio": "^8.20.4",
         "webpack": "^5.88.2",
         "webpack-cli": "^5.1.4"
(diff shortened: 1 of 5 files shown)


sourishkrout:
Not quite `SevereServiceError: Couldn't set up Chromedriver Response code 404 (Not Found)`.

sourishkrout:
As per https://raw.githubusercontent.com/microsoft/vscode/1.86.0/cgmanifest.json it requires chromedriver 118.

Weirdly enough the chromedriver is not available yet: https://chromedriver.storage.googleapis.com/LATEST_RELEASE_118

sourishkrout:
@pastuxso it appears the way Electron/Chromium is being matched with Chromedriver has changed: https://chromedriver.chromium.org/downloads/version-selection

It appears that there is no Chromedriver for version 118 specifically. The doc above says to use the previous version, so we may have to implement a retry that incrementally decreases the version number. Something we should probably "stage" locally to prove it works before implementing all the logic in the upstream project:

https://github.com/webdriverio-community/wdio-vscode-service/blob/main/src/launcher.ts#L206-L238

pastuxso (approved):
LGTM :rocket:

PR #1094, opened 2024-02-03
Refactor executor

sourishkrout:
Doing some spring cleaning of long-overdue code that has grown organically. It's not where I want it to be but quite a step forward. The primary goal is to provide refactor the baseline of exectors/runners to couple with interfaces vs implementations before introducing upcoming functionality.


sourishkrout:
@pastuxso please review. Also, we might want to consider lowering the coverage threshold to 80% until we have time to bring it back up to 85%.

pastuxso:
@sourishkrout :white_check_mark: done.

pastuxso (approved):

PR #1116, opened 2024-02-12
fetch APIs are part of core since node 18

sourishkrout:
Removing superfluous dependency now that `undici` is part of core.


sourishkrout:
Please review, @pastuxso. Should be uncontroversial.

PR #1121, opened 2024-02-15
Refactor platform API message handling in kernel.ts

pastuxso:
- Remapped cloud API requests to platform API requests when platform authentication is enabled.
- Updated the response messages for saving cell executions to use platform API response messages.
- Added tests for platform API request handling and saving cell executions.


sourishkrout (approved):
👍 LGTM - Using an TS interface ViewModel could also help an additional strategy to keep the "backend" interchangeable. However, if we are planning to decom the old backend it's probably not worth doing all the refactoring to implement said strategy.

Let's merge.

PR #1137, opened 2024-02-19
Fixes broken panels using current GH Auth

pastuxso:
During the implementation of Platform Authentication, an issue was introduced affecting the current authentication flow. The panels are failing to load as a result of this issue


sourishkrout (approved):
👍 LGTM

PR #1147, opened 2024-02-21
Add GCE VM Instances view

degrammer:
Screenshot 2024-02-21 at 1 22 06 PM


sourishkrout (approved):
👍 LGTM. WE should ear-mark that we should add a GCE instance detail view, too.

PR #1148, opened 2024-02-22
Restores functionality for Stateful IDP support on localhost

pastuxso:
These adjustments guarantee the correct operation of Stateful IDP on localhost. They involve the removal of mappings associated with specific subdomains linked to previous platform FQDNs.


sourishkrout (approved):
👍 LGTM

PR #1152, opened 2024-02-22
Remove throw if there is not a Stateful session

peraltafederico:
Remove the error throw if there is no a Stateful session. This way the "EMPTY" string will be sent as a token to the webview panels app, so it can handle unauthenticated users.


pastuxso (approved):
LGTM!

PR #1154, opened 2024-02-24
Refactor Stateful IDP to use Platform endpoints

pastuxso:
Refactor the Stateful Identity Provider to remove the settings from the extension and delegate it to the API.

Things to do in the next iterations:

- Enhance the security between the extension and the Platform API,TLS Auth maybe...
- Increase the test coverage

### DIFF
--- a/__mocks__/vscode.ts
+++ b/__mocks__/vscode.ts
@@ -4,8 +4,11 @@ import { join } from 'node:path'
 
 import { vi } from 'vitest'
 import { URI } from 'vscode-uri'
-import type { Command, Disposable } from 'vscode'
+import { Command, Disposable as DisposableOriginal } from 'vscode'
 
+export const Disposable = {
+  from: vi.fn()
+}
 
 export const notebooks = {
   createNotebookController: vi.fn().mockReturnValue({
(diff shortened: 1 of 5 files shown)


sourishkrout (approved):
👍 LGTM.

sourishkrout:
> * Enhance the security between the extension and the Platform API,TLS Auth maybe...

If we make sure we only have allow URLs using TLS (URI with `https` scheme) we basically have all the benefits of TLS authentication and encryption, no? Not sure I'm following here.

pastuxso:
> > * Enhance the security between the extension and the Platform API,TLS Auth maybe...
> 
> If we make sure we only have allow URLs using TLS (URI with `https` scheme) we basically have all the benefits of TLS authentication and encryption, no? Not sure I'm following here.

I mean something like Client Certificate Authentication.

sourishkrout:
> > > * Enhance the security between the extension and the Platform API,TLS Auth maybe...
> > 
> > 
> > If we make sure we only have allow URLs using TLS (URI with `https` scheme) we basically have all the benefits of TLS authentication and encryption, no? Not sure I'm following here.
> 
> I mean something like Client Certificate Authentication.

Gotcha.

PR #1166, opened 2024-02-26
Introduce notebook panels (bottom by default in IDE)

sourishkrout:
New panel for development. It's a placeholder. Will be removed in both edge/stable releases for now until fully completed.

image


sourishkrout:
@degrammer please review. introducing a new panel at the bottom with a placeholder for now.

PR #1187, opened 2024-03-05
Avoid back pressure

sourishkrout:
Grab all emissions as a single array, then sort. As opposed to sorting every time a new emission shows up.

This fixes a significant slowdown in the extension host for repos with lots of markdown files.


pastuxso (approved):
LGTM :rocket:

PR #1213, opened 2024-03-15
Fix env store prompts and remove tooltip for env store value

degrammer:
- Remove tooltip from env store value
- Don't remember Run with prompts after cell execution


sourishkrout (approved):
👍 Works

PR #1218, opened 2024-03-15
Send fileName when saving a cell

peraltafederico:
Send notebook fileName when saving a cell output for Stateful Platform


pastuxso (approved):
LGTM :rocket:

PR #1242, opened 2024-03-29
Integrate unArchiveCellExecution 

peraltafederico:
Integrate `unArchiveCellExecution` so users can restore cells when using the Cloud Panels. It also emits the "onCell" events once the archiving/restoring is done instead of when it was started, this way the Panel App knows when the mutation has finished.


sourishkrout (approved):
👍 LGTM

PR #1244, opened 2024-04-01
Fix using the wrong user session for auto-save

degrammer:
This PR fixes using the wrong user session when platform is enabled.


sourishkrout (approved):
👍 LGTM

PR #1245, opened 2024-04-01
Introducing exploring/authoring mode

degrammer:
Fixes https://github.com/stateful/vscode-runme/issues/1168

The exploring mode is the default mode. This will be available at Notebook level, the mode will be persisted across notebooks. 

explorer-mode

### DIFF
--- a/package.json
+++ b/package.json
@@ -416,7 +416,6 @@
           "group": "navigation"
         },
         {
-<<<<<<< HEAD
           "command": "runme.notebookGistShare",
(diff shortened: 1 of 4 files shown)


sourishkrout (approved):
Let's temporarily make the mode that add new cells default and set the `when` condition on the two button states to `false` so they are not displayed, then merge. I want to take a bit more time as we work on the cloud renderers to figure out naming and UX. However, not letting this branch go stale and making it mergable is the way to go.

sourishkrout (approved):
👍 LGTM

Issue #1235, opened 2024-03-26
Auto-save extension default overrides user session selection


joseeantonior:
The auto-save default set in the extension setting overrides the user selection during the particular session. The user's selection should stick until the window is closed or changed by the user, and not be overridden by the extension setting. We should start thinking of extension settings as presets, but not hard decisions.

PR #1247, opened 2024-04-02
Keep auto-save user selection across notebooks

degrammer:
Introducing a fix to respect user selection when using auto-save feature.
Currently it uses the extension default when opening a Notebook, overwriting the user selection.

This PR fixes https://github.com/stateful/vscode-runme/issues/1235


sourishkrout (approved):
👍 LGTM

PR #1248, opened 2024-04-02
Rely on status instead of spec for the owl store

degrammer:
Moving from spec to status (and custom typing) to keep it as the source of truth for the owl store.

# How to test?
Create a .env.spec with the following values:

```JWT_SECRET= # Secret!
ANON_KEY=  # Secret!
POSTGRES_PASSWORD=  # Password!
DASHBOARD_USERNAME= # Plain!
DASHBOARD_PASSWORD=  # Password!
```

Add values to an .env file and check if the env store reports correctly the missing values, the copy / eye buttons appears acoordingly.

Screenshot 2024-04-02 at 1 28 48 PM

### DIFF
--- a/src/client/components/env/store.ts
+++ b/src/client/components/env/store.ts
@@ -107,7 +107,7 @@ export default class Table extends LitElement {
             case 'createdAt':
               return html`${row.createdAt ? formatDateWithTimeAgo(new Date(row.createdAt)) : ''}`
             case 'updatedAt':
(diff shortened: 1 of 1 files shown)


sourishkrout (changes requested):
LGTM 👍.

@degrammer would it be possible to give the `Created` column a relative time formatting, e.g. `2 min ago`?

sourishkrout (changes requested):
Only `Created` not `Updated`. We want the full timestamp on updated please, @degrammer.

sourishkrout (approved):
👍 LGTM. Let's merge.

PR #1250, opened 2024-04-03
platform gif padding

admc:
Add some padding to the platform gif so the stateful logo isn't massive in the file and extension views


sourishkrout (approved):
👍 LGTM

PR #1256, opened 2024-04-04
Use extension env var collection API for "globals vars"

sourishkrout:
Two things:

1. This will simplify setting Runme-related of ENV vars.
2. Moves runner v1 types into a namespace to prep v2 migration.


pastuxso (approved):
LGTM :+1:

PR #1260, opened 2024-04-06
Fix proto integration

sourishkrout:
No longer use Javascript generation since appears broken upstream.


sourishkrout:
Turns out this "fixable" since it's likely a bug in https://github.com/timostamm/protobuf-ts/issues/643. TBD.

PR #1267, opened 2024-04-09
Remove vercel, fresh, enhance examples, move folders used for tests to fixtures

degrammer:


sourishkrout (changes requested):
@degrammer no deprecation notices? I'd suggest to keep the sub-directories and their READMEs around with something like we've put on top of https://github.com/stateful/vscode-runme/blob/main/examples/branchGPT.md. Except let's delete the rest of the content. Btw, let's also do that for `branchGPT.md` while we're at it.

sourishkrout:
> > @degrammer no deprecation notices? I'd suggest to keep the sub-directories and their READMEs around with something like we've put on top of https://github.com/stateful/vscode-runme/blob/main/examples/branchGPT.md. Except let's delete the rest of the content. Btw, let's also do that for `branchGPT.md` while we're at it.
> 
> @sourishkrout Done, added deprecation notice and removed associated contents

Also remove the content in the notebook please, @degrammer. If there's nothing to run anymore, it makes no sense to keep anything past the deprecation notice.

sourishkrout (changes requested):
See comment please.

joseeantonior:
LGTM.

PR #1272, opened 2024-04-12
Fix styles for fx terminal and add more examples to the README file

degrammer:
This PR fixes some style issues when using specific themes for fx JSON viewer, it also adds another example usage for interacting with a kubernetes resources list.

### DIFF
--- a/src/client/components/terminal/index.ts
+++ b/src/client/components/terminal/index.ts
@@ -305,6 +305,12 @@ export class TerminalView extends LitElement {
     #terminal {
       position: relative;
     }
+
+    /** xf styles */
+    .xterm-bg-257,
+    .xterm-bg-7 {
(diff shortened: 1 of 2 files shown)


src/client/components/terminal/index.ts
```diff
@@ -305,6 +305,12 @@ export class TerminalView extends LitElement {
     #terminal {
       position: relative;
     }
+
+    /** xf styles */
+    .xterm-bg-257,
+    .xterm-bg-7 {
+      background-color: var(--vscode-editor-selectionBackground) !important;
```
sourishkrout, line 312: @degrammer will we have to worry about any unintended side-effects introducing these two color mappings?

sourishkrout (changes requested):
Just a question.

PR #1289, opened 2024-04-18
Allow to save session output in Gist

degrammer:
This PR allows to store a session output as a Gist. 
It also introduces a new **GitHubServiceFactory** that allows to use the existing Octokit API in a more flexible way.


sourishkrout (changes requested):
It feels odd that these GitHub classes/functions are nested under `executors` when they don't have anything to do with Kernel executors (being run with the ▶️ button). I'd suggest to find a less confusing home for them and have the `workflows.ts` (which I believe is a kernel executor) depend on it.

Also, please see other comments, @degrammer.

sourishkrout (approved):
👍 LGTM

PR #1302, opened 2024-04-22
Add python example

degrammer:
Now Python interpreter detection is fixed (see https://github.com/stateful/runme/issues/552), we're adding back the Python example for Shebang.

Screenshot 2024-04-22 at 12 43 07 PM


sourishkrout (approved):
Works! 👍 LGTM

PR #1305, opened 2024-04-23
Enable GH action tests

degrammer:


pastuxso (approved):

sourishkrout:
@degrammer what was the problem? Can you point out in the description so we have a historical record?

PR #1321, opened 2024-05-03
Fix test workflow file

degrammer:
The test workflow file included some settings that are not available in this repository, it was causing the test to fail once merged in main.


sourishkrout (approved):
Works for me as long as we get a ✅ build

PR #1323, opened 2024-05-03
Only require cell lifecycle identity by default

sourishkrout:
The group of users who require this is much smaller than what makes sense for default now.


sourishkrout:
Please review @pastuxso.

pastuxso (approved):
LGTM :+1: 

image

Issue #1304, opened 2024-04-22
Visual feedback for `shebang++`


sourishkrout:
If a cell name conforms to the regex below, Runme will store its `stdout` under the given name.

https://github.com/stateful/runme/blob/main/internal/runner/service.go#L46-L47

Explore if it's possible to provide visual feedback in a cell's status bar where the name is displayed. Perhaps, names that don't match could have a muted-gray whereas the ones matching could have a muted-green.

https://gist.github.com/sourishkrout/892e83e633aecf2283113030b24686ae#file-outputs-01HW3HF36Y2HVA4XRW73ZZQZEV.md

UX for https://github.com/stateful/vscode-runme/pull/1303

PR #1334, opened 2024-05-06
Add shebang visual feedback

degrammer:
Discovery work for https://github.com/stateful/vscode-runme/issues/1304

It wasn't possible to change the color of the Cell Status Bar, hence, changing the Icon + Tooltip to reflect the state of it.

shebang++


sourishkrout:
> It wasn't possible to change the color of the Cell Status Bar, hence, changing the Icon + Tooltip to reflect the state of it.

That's fair, can we use a different icon from than "x" though? Seems a little too alarmist 🚨.

sourishkrout (changes requested):
Let's change the icon. If unclear which one let's discuss options on Slack.

sourishkrout (approved):
👍 LGTM

Issue #1307, opened 2024-04-24
Rewire "Save" and add "Gist" buttons on cells


sourishkrout:
## Save button 
Instead of showing a modal with "Secure collaboration ... the Stateful Platform. Click "Open" to learn more." when `runme.app.enableShare` is turned off, let's use the setting to control it's visibility only. In other words, if it's turned off, don't show the button. If it's turned on, attempt to save to cloud platform.

## Gist button 
Open the session outputs file (akin to "Session Outputs" button), if none is available (e.g. `Auto-Save` is `off`) show modal to remind people to turn on Auto-Save first.

<img width="959" alt="image" src="https://github.com/stateful/vscode-runme/assets/250527/548861bd-f7b2-4541-ae2a-14a4ef48f08e">

joseeantonior:
Notes: Have PR ready but don't merge until reviewed by Sebastian. Approval by Sebastian is required before a merge.

PR #1335, opened 2024-05-08
Add Cell Gist Button

degrammer:
Introducing Gist button for Cell ouputs

Fixes https://github.com/stateful/vscode-runme/issues/1307

gist-cell

This PR also organizes the StatusBarItems for elements that are displayed only for Notebooks vs Session Output only.


sourishkrout (approved):
LGTM 👍

PR #1356, opened 2024-05-16
Generate cells using Foyle AI

jlewi:
* Create a command which sends an RPC to Foyle to generate cells
* Resulting cells are inserted into the notebook

Related to stateful/runme#574

### DIFF
--- a/package.json
+++ b/package.json
@@ -272,6 +272,10 @@
       {
         "command": "runme.cellGistShare",
         "title": "Click to generate GitHub gist link"
+      },
+      {
+        "command": "runme.aiGenerate",
+        "title": "Generate cells using the Foyle AI assistant"
       }
     ],
     "menus": {
--- a/package.json
+++ b/package.json
@@ -1014,6 +1014,12 @@
         "key": "L",
         "when": "notebookEditorFocused && !inputFocus && !notebookOutputInputFocused",
         "command": "runme.keybinding.noop"
+      },
+      {
+        "command": "runme.aiGenerate",
+        "key": "win+;",
+        "mac": "cmd+;",
+        "when": "resourceExtname==.md || resourceExtname==.markdown"
       }
     ],
     "terminal": {
--- a/package.json
+++ b/package.json
@@ -804,6 +804,11 @@
             "scope": "window",
             "default": false,
             "markdownDescription": "If set to `true` enables Stateful Authentication Provider"
+          },
+          "runme.foyleAddress": {
+            "type": "string",
+            "default": "localhost:9080",
+            "description": "The address of the Foyle Agent grpc service. The agent generates completions. This should in the form {host}:{port}."
           }
         }
       }
--- a/src/extension/ai/client.ts
+++ b/src/extension/ai/client.ts
@@ -1,0 +1,12 @@
+import { AIServiceClient } from '@buf/stateful_runme.community_timostamm-protobuf-ts/runme/ai/v1alpha1/ai_pb.client'
+import { GrpcTransport } from '@protobuf-ts/grpc-transport'
+import { ChannelCredentials } from '@grpc/grpc-js' // Import the missing module
+
+// TODO(jeremy): Should we take transport as an argument?
+export function initAIServiceClient(): AIServiceClient {
+  const transport = new GrpcTransport({
+    host: 'localhost:9080',
+    channelCredentials: ChannelCredentials.createInsecure(),
+  })
+  return new AIServiceClient(transport)
+}
--- a/src/extension/ai/client.ts
+++ b/src/extension/ai/client.ts
@@ -1,11 +1,22 @@
 import { AIServiceClient } from '@buf/stateful_runme.community_timostamm-protobuf-ts/runme/ai/v1alpha1/ai_pb.client'
 import { GrpcTransport } from '@protobuf-ts/grpc-transport'
 import { ChannelCredentials } from '@grpc/grpc-js' // Import the missing module
+import * as vscode from 'vscode'
+
+// extName is the name of the extension.
+// This needs to match the value in package.json
+// It is used to look up extension configuration
+export const extName = 'runme'
 
 // TODO(jeremy): Should we take transport as an argument?
 export function initAIServiceClient(): AIServiceClient {
+  const config = vscode.workspace.getConfiguration(extName)
+  // Include a default so that address is always well defined
+  const address = config.get<string>('foyleAddress', 'localhost:9080')
+
+  console.log(`Connecting to Foyle GRPC: ${address}`)
   const transport = new GrpcTransport({
-    host: 'localhost:9080',
+    host: address,
     channelCredentials: ChannelCredentials.createInsecure(),
   })
   return new AIServiceClient(transport)
--- a/src/extension/ai/converters.ts
+++ b/src/extension/ai/converters.ts
@@ -1,0 +1,86 @@
+// Conversion routines between the protocol buffer format and VSCode's representation of a notebook
+//
+// See ../vscode_apis.md for an exlanation. It is very helpful for understanding this folder.
+
+import * as vscode from 'vscode'
+//import * as metadata from './metadata'
+//import * as constants from './constants'
+import {
+  Notebook,
+  Cell,
+  CellKind,
+  CellOutput,
+  CellOutputItem,
+} from '@buf/stateful_runme.community_timostamm-protobuf-ts/runme/parser/v1/parser_pb'
+
+// cellToCellData converts a NotebookCell to a NotebookCellData.
+// NotebookCell is an interface used by the editor.
+// NotebookCellData is a concrete class.
+export function cellToCellData(cell: vscode.NotebookCell): vscode.NotebookCellData {
+  let data = new vscode.NotebookCellData(
+    cell.kind,
+    cell.document.getText(),
+    cell.document.languageId,
+  )
+
+  data.metadata = cell.metadata
+  data.outputs = []
+  for (let o of cell.outputs) {
+    data.outputs.push(o)
+  }
+  return data
+}
+
+// cellDataToProto converts an instance of NotebookCellData to the proto.
+export function cellDataToProto(cell: vscode.NotebookCellData): Cell {
+  let cellPb = Cell.create()
+  cellPb.value = cell.value
+  cellPb.languageId = cell.languageId
+  if (cell.kind === vscode.NotebookCellKind.Code) {
+    cellPb.kind = CellKind.CODE
+  } else {
+    cellPb.kind = CellKind.MARKUP
+  }
+
+  // TODO(jeremy):  cell.metadata is type map[string]Any; cellPb.metadata is type map[string]string
+  // How do we handle the conversion?
+  //cellPb.metadata = cell.metadata
+
+  let cellOutputs: vscode.NotebookCellOutput[] = []
+  if (cell.outputs !== undefined) {
+    cellOutputs = cell.outputs
+  }
+  for (const output of cellOutputs) {
+    let out = CellOutput.create()
+    out.items = []
+    for (const item of output.items) {
+      let outItem = CellOutputItem.create()
+      outItem.data = item.data
+      outItem.mime = item.mime
+      out.items.push(outItem)
+    }
+    cellPb.outputs.push(out)
+  }
+  return cellPb
+}
+
+// protoToCellData converts a cell proto to a NotebookCellData.
+export function protoToCellData(cellPb: Cell): vscode.NotebookCellData {
+  let kind = vscode.NotebookCellKind.Markup
+  if (cellPb.kind === CellKind.CODE) {
+    kind = vscode.NotebookCellKind.Code
+  }
+
+  let newCell = new vscode.NotebookCellData(kind, cellPb.value, cellPb.languageId)
+
+  newCell.metadata = cellPb.metadata
+  newCell.outputs = []
+  for (let output of cellPb.outputs) {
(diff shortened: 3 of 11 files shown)


src/extension/ai/converters.ts
```diff
@@ -0,0 +1,86 @@
...
+  Cell,
+  CellKind,
+  CellOutput,
+  CellOutputItem,
+} from '@buf/stateful_runme.community_timostamm-protobuf-ts/runme/parser/v1/parser_pb'
+
+// cellToCellData converts a NotebookCell to a NotebookCellData.
+// NotebookCell is an interface used by the editor.
+// NotebookCellData is a concrete class.
+export function cellToCellData(cell: vscode.NotebookCell): vscode.NotebookCellData {
```
jlewi, line 19: @sourishkrout Do I need these conversion routines to convert between the protos and vscode's native data structures? Are there existing routines I can reuse?

src/extension/ai/converters.ts
```diff
@@ -0,0 +1,86 @@
...
+  let cellPb = Cell.create()
+  cellPb.value = cell.value
+  cellPb.languageId = cell.languageId
+  if (cell.kind === vscode.NotebookCellKind.Code) {
+    cellPb.kind = CellKind.CODE
+  } else {
+    cellPb.kind = CellKind.MARKUP
+  }
+
+  // TODO(jeremy):  cell.metadata is type map[string]Any; cellPb.metadata is type map[string]string
```
jlewi, line 45: @sourishkrout Is there existing conversion code I can use here? If not is there a way to do this other than just testing if the values are strings and if not ignoring them?

tests/extension/ai/converters.test.ts
```diff
@@ -0,0 +1,30 @@
...
+// TODO(jeremy): Importing vscode here appears to cause this to fail when run as a unit-test using
+// npx runme run test:unit
+import * as vscode from 'vscode'
+import { describe, it } from 'vitest'
+import { GenerateCellsRequest } from '@buf/stateful_runme.community_timostamm-protobuf-ts/runme/ai/v1alpha1/ai_pb'
+import { CellKind } from '@buf/stateful_runme.community_timostamm-protobuf-ts/runme/parser/v1/parser_pb'
+
+import * as converters from '../../../src/extension/ai/converters'
+
+// TODO(jeremy): Add more tests assuming we do in fact need the converters.
```
jlewi, line 12: @sourishkrout I didn't flush out the unittests for the converters because I wasn't sure we actually needed them.  I wasn't quite sure how to structure the test if we do. It looks like importing vscode means we can no longer treat this as a unittest.

jlewi:
@sourishkrout Would you PTAL? The code is a bit rough (lots of lint errors); however I had some questions about how various things should be done and I thought you might be able to provide some pointers?

FWIW the code does work; I was able to run it and get suggestions provided by Foyle

package.json
```diff
@@ -1010,6 +1019,12 @@
         "key": "L",
         "when": "notebookEditorFocused && !inputFocus && !notebookOutputInputFocused",
         "command": "runme.keybinding.noop"
+      },
+      {
+        "command": "runme.aiGenerate",
+        "key": "win+;",
+        "mac": "cmd+;",
+        "when": "resourceExtname==.md || resourceExtname==.markdown"
```
sourishkrout, line 1027: `notebookType == runme` might be a better choice here instead of keying on file extension

src/extension/ai/client.ts
```diff
@@ -0,0 +1,23 @@
...
+import * as vscode from 'vscode'
+
+// extName is the name of the extension.
+// This needs to match the value in package.json
+// It is used to look up extension configuration
+export const extName = 'runme'
+
+// TODO(jeremy): Should we take transport as an argument?
+export function initAIServiceClient(): AIServiceClient {
+  const config = vscode.workspace.getConfiguration(extName)
```
sourishkrout, line 13: nit, in runme-land we have a few zod-schemata-based abstractions to handle config. not super important right now.

src/extension/ai/converters.ts
sourishkrout: I believe this static method is what you're looking for:

https://github.com/stateful/vscode-runme/blob/main/src/extension/serializer.ts#L740-L757

src/extension/ai/converters.ts
sourishkrout: this static method should do the job: https://github.com/stateful/vscode-runme/blob/main/src/extension/serializer.ts#L302-L333

src/extension/ai/generate.ts
```diff
@@ -0,0 +1,85 @@
...
+    req.notebook.cells.push(cellPb)
+  }
+
+  let client = initAIServiceClient()
+
+  client
+    .generateCells(req)
+    .then((finished) => {
+      let response = finished.response
+      // TODO(jeremy): We should have the server add the traceId to the response and then we should
```
sourishkrout, line 50: There's a logger abstraction that makes logging human-friendly with most of the bells and whistles, e.g. here https://github.com/stateful/vscode-runme/blob/main/src/extension/server/runmeServer.ts#L38

tests/extension/ai/converters.test.ts
sourishkrout: 👍  assuming you can reuse the Serializer's marshal and revive methods you should be good to forgo this.

src/extension/ai/client.ts
```diff
@@ -0,0 +1,23 @@
...
+import { ChannelCredentials } from '@grpc/grpc-js' // Import the missing module
+import * as vscode from 'vscode'
+
+// extName is the name of the extension.
+// This needs to match the value in package.json
+// It is used to look up extension configuration
+export const extName = 'runme'
+
+// TODO(jeremy): Should we take transport as an argument?
+export function initAIServiceClient(): AIServiceClient {
```
sourishkrout, line 12: we've been keeping the `gRPC` related abstractions under https://github.com/stateful/vscode-runme/tree/main/src/extension/grpc. Including re-exporting the generated protos, e.g. https://github.com/stateful/vscode-runme/blob/main/src/extension/grpc/projectTypes.ts.

This allows us to avoid spreading these long imports `'@buf/stateful_runme.community_timostamm-protobuf-ts/runme/ai/v1alpha1/ai_pb.client'` across the code base and helps with API version migration down the road.

Can always do this later.

package.json
```diff
@@ -800,6 +804,11 @@
             "scope": "window",
             "default": false,
             "markdownDescription": "If set to `true` enables Stateful Authentication Provider"
+          },
+          "runme.foyleAddress": {
```
sourishkrout, line 808: I'm wondering if we should have a feature flag for Foyle for the time being. We do have a simple abstraction that has served us well:

https://github.com/stateful/vscode-runme/blob/main/src/extension/kernel.ts#L179-L181

src/extension/kernel.ts
```diff
@@ -157,16 +157,22 @@ export class Kernel implements Disposable {
...
-      workspace.onDidOpenNotebookDocument(this.#handleOpenNotebook.bind(this)),
-      workspace.onDidSaveNotebookDocument(this.#handleSaveNotebook.bind(this)),
-      window.onDidChangeActiveColorTheme(this.#handleActiveColorThemeMessage.bind(this)),
-      window.onDidChangeActiveNotebookEditor(this.#handleActiveNotebook.bind(this)),
-      this.panelManager,
-      this.onVarsChangeEvent,
-      this.registerTerminalProfile(),
-    )
+
+    // DO NOT COMMIT this was a hack because I was getting crashes when trying to run the extension
```
sourishkrout, line 161: let's figure out what's going on - sooner or later the lack of running these handlers will rear it's ugly head

sourishkrout (changes requested):
Overall good stuff. Filled in some gaps for you. Hopefully the requested changes are reasonable to get this merged.

I think it's probably wise to forgo unit tests with mocking for the app/UI logic and e2e integration tests until the integration moves out of alpha.

src/extension/kernel.ts
jlewi: Seems to be working for me now. I have no idea what changed. I uninstalled the extension and reinstalled it as part of trying to install a dev version of the extension in my vscode; so maybe it was that 🤷

src/extension/ai/client.ts
jlewi: Done.

jlewi (commented):
Thanks @sourishkrout . I believe I've addressed all your comments with the exception of the zota config issue and the experiments flag; see my replies to those comments.

src/extension/ai/converters.ts
sourishkrout: @jlewi is there maybe a file missing in the latest commits?

image

src/extension/ai/client.ts
sourishkrout: it's all in `src/utils/configuration.ts` but we ok changing it post-merge

src/extension/ai/generate.ts
sourishkrout: There's only a simple `runme.server.enableLogger` bool switch that will enable debug logging to `Output->Runme`.

package.json
sourishkrout: > What would the experiment flag control? Is there a way to gate the configuration options based on the experiment? I think an experiment flag might be redundant with foyleAddress. I lean towards using a non-empty value of foyleAddress a good way to signal foyle is enabled. I think if we tell users they need to do two things - 1) set a non-empty address 2) flip the experiment - it increases the likelihood of mistakes due to user error; i.e. user only sets one of them and wonders why it isn't working.

non-empty `foyleAddress` would work sufficiently. the experiment/feature flag would 1.) allow explicit opt-it and 2.) special case anywhere in the extension code accordingly. Specifically for logging detailed execution data we'd like to disable persistent logs for any unsuspecting Runme user. I agree we can defer making those change as part of resolving Foyle-compatible logging.

sourishkrout (changes requested):
Won't compile/build locally right now... likely due to a missing file? See comments.

Otherwise 👍 .

src/extension/ai/converters.ts
jlewi: @sourishkrout Can you help me figure out what's going on here and how to fix it.  I think it should have been

```
import {Serializer} from '../../types'
```

I'm trying to import [Notebook defined here](https://github.com/stateful/vscode-runme/blob/497053bb5b147b0ceb4a1d9b9cc62c925cd66543/src/types.ts#L63).

I'm trying to instantiate an instance of Notebook to pass to Notebook.Revive here
https://github.com/stateful/vscode-runme/pull/1356/files#diff-7561d1e499859730c4937494e85d7dddfbe11332a496476afc5d082860df7335R81

What's the proper way

1. Instantiate a notebook
2. Populate the cells in the notebook with the proto

I'm a little confused by what's going on. I see the error in vscode but ` npm run bundle` succeeds and I thought I had installed it successfully from source.

jlewi: So I think I partially understand the error. Serializer.Notebook is a type not a class so I can't use new.  I wrote some conversion code to convert the proto type to the Serializer.CellType

PTAL

jlewi:
@sourishkrout Sorry about that. I think its fixed.

src/extension/ai/converters.ts
sourishkrout: Correct. The `Serializer.Notebook` is a type that in structural typing sense fits different Parsers. Runme has a legacy WASM-based parser from its prototype days. Fixing the import path fixed the error for me.

I have seen the Typescript LSP and/or the ESlint LSP getting "confused" which happens when files are being moved around right under it. The easy fix is to restart them using the command palette. In any case, it compiles cleanly now on all fronts webpack/tsc, and eslint & typescript LSPs.

src/extension/ai/generate.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    if (cell.kind === CellKind.CODE) {
+      kind = vscode.NotebookCellKind.Code
+    }
+
+    let newCell: Serializer.Cell = {
+      value: cell.value,
+      metadata: cell.metadata,
+      kind: kind,
+      // TODO(jeremy): Should we include outputs? The generate response should never contain outputs so we shouldn't
```
sourishkrout, line 96: fyi, outputs are entirely optional in Runme. We only need them when recording of Session files is enabled which writes to a "carbon copy" of the original markdown notebook; never back to the original source file.

sourishkrout (approved):
👍 LGTM

sourishkrout:
One of the unit tests failed which is an easy fix. `registerCommand` is called an additional time for the Foyle/AI cmd.

```
 FAIL  tests/extension/extension.test.ts > initializes all providers
AssertionError: expected "spy" to be called 39 times, but got 40 times
 ❯ tests/extension/extension.test.ts:102:36
    100|   expect(notebooks.registerNotebookCellStatusBarItemProvider).toBeCall…
    101|   expect(workspace.registerNotebookSerializer).toBeCalledTimes(1)
    102|   expect(commands.registerCommand).toBeCalledTimes(39)
       |                                    ^
    103|   expect(window.registerTreeDataProvider).toBeCalledTimes(1)
    104|   expect(window.registerUriHandler).toBeCalledTimes(1)

Error: AssertionError: expected "spy" to be called 39 times, but got 40 times
 ❯ tests/extension/extension.test.ts:102:36
```

jlewi:
Thank you @sourishkrout ; I fixed the test.

sourishkrout:
Merging. - The e2e integration test's failing because GitHub Actions are being finicky and have been under the weather with outages the last few days. It's not broken.

PR #1358, opened 2024-05-17
Updates NODE_OPTIONS to avoid deployment issues

pastuxso:
When the **NODE_OPTIONS** is already set in the GitHub actions file **runme** omits prompting, and it's resolved from the GitHub Action workflow.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -69,7 +69,7 @@ jobs:
         run: npm run build:prod
         env:
           NODE_ENV: production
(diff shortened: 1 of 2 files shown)


sourishkrout:
Closing since https://github.com/stateful/vscode-runme/commit/f6299b33a168d59b01e5df4937fc942e9039f0bf accomplishes the same. However, having less runtime-specific stuff in the YAML is desirable because only Actions run the YAML and it creates these snowflake issues.

Issue #1354, opened 2024-05-15
Invert persistent vs ephemeral cell IDs


sourishkrout:
Sort of revert what we did in this PR: https://github.com/stateful/vscode-runme/pull/938

And then salvaged in this PR: https://github.com/stateful/vscode-runme/pull/1252

The remaining problem right now is that when cell IDs are turned off we have to remove them from the document. *By having the core logic rely on the ephemeral IDs* and make them the same as the persistent ones (already the case). Cell IDs will be optional again no matter whether they are persistently present or not.

sourishkrout:
When identity is turned off for cells, saving this file should write exactly the same, as in don't add or remove any IDs from cells.

![Image](https://github.com/stateful/vscode-runme/assets/250527/4600a8ba-546b-4881-8118-8d964dad7a8e)

In the current implementation it will either add IDs to **all** cells or remove them from *all* cells.

PR #1359, opened 2024-05-17
Fix: Preserve identity IDs when IDT is disabled

pastuxso:
The issue is that when IDT is turned off, identity IDs need to remain unchanged, even if they are included in a notebook.

This includes:

- **runme.dev/stageId**: This is a new metadata field used to transfer the IDT during serialize/deserialize requests only when **RunmeIdentity** is **UNSPECIFIED**.

- Updates to **assertDocumentContains** to only match a portion of the document.

- Test updates

Closes: #1354

### DIFF
--- a/src/extension/serializer.ts
+++ b/src/extension/serializer.ts
@@ -355,8 +355,8 @@ export abstract class SerializerBase implements NotebookSerializer, Disposable {
     }
 
     data.cells.forEach((cell) => {
-      if (identity === RunmeIdentity.UNSPECIFIED && cell.metadata?.['runme.dev/stageId']) {
-        cell.metadata['id'] = cell.metadata['runme.dev/stageId']
+      if (identity === RunmeIdentity.UNSPECIFIED && cell.metadata?.['runme.dev/originalId']) {
+        cell.metadata['id'] = cell.metadata['runme.dev/originalId']
         return
       }
 
@@ -609,7 +609,7 @@ export class GrpcSerializer extends SerializerBase {
             return
           }
           if (cell.metadata?.['id']) {
-            cell.metadata['runme.dev/stageId'] = cell.metadata['id']
+            cell.metadata['runme.dev/originalId'] = cell.metadata['id']
           }
         })
         break
--- a/src/extension/serializer.ts
+++ b/src/extension/serializer.ts
@@ -355,6 +355,11 @@ export abstract class SerializerBase implements NotebookSerializer, Disposable {
     }
 
     data.cells.forEach((cell) => {
+      if (identity === RunmeIdentity.UNSPECIFIED && cell.metadata?.['runme.dev/stageId']) {
+        cell.metadata['id'] = cell.metadata['runme.dev/stageId']
+        return
+      }
+
       if (cell.metadata?.['id'] !== undefined) {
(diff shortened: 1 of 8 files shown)


sourishkrout:
@degrammer @pastuxso I'll be doing the review on this. Please don't **request reviews from one the more people at once** or jump in unprompted. Context and duplication of work is costly.

src/extension/serializer.ts
```diff
@@ -353,6 +353,11 @@ export abstract class SerializerBase implements NotebookSerializer, Disposable {
     }
 
     data.cells.forEach((cell) => {
+      if (identity === RunmeIdentity.UNSPECIFIED && cell.metadata?.['runme.dev/stageId']) {
```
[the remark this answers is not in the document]
pastuxso: @degrammer During the deserialization/serialization/save process, the **stageId** is used to transport the ID and then recover it to its original form as the ID. Instead of "stage" it could be called a "snapshot," "copy," or "tmp," which sounds clearer to me. However, I will explore finding a new, more descriptive field name.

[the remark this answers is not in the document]
pastuxso: @degrammer, I updated it to "originalId".

src/extension/serializer.ts
```diff
@@ -596,9 +601,18 @@ export class GrpcSerializer extends SerializerBase {
...
-      case RunmeIdentity.UNSPECIFIED:
       case RunmeIdentity.DOCUMENT:
         break
+      case RunmeIdentity.UNSPECIFIED:
+        data.cells.forEach((cell) => {
+          if (cell.kind !== CellKind.CODE) {
+            return
+          }
+          if (cell.metadata?.['id']) {
+            cell.metadata['runme.dev/originalId'] = cell.metadata['id']
```
sourishkrout, line 612: Don't think this is the right solution. It tries to fix "the issue" at the tail end of serialization instead of deserialization where it seems to originate.

I believe this code here's not "smart enough" yet. It ignores both identity requirements and the fact that `runme.dev/id` is not set which means cell identity wasn't required in the first place.
https://github.com/stateful/vscode-runme/blob/lifecycle-id-fixes/src/extension/serializer.ts#L319-L324

src/extension/serializer.ts
```diff
@@ -355,6 +355,11 @@ export abstract class SerializerBase implements NotebookSerializer, Disposable {
     }
 
     data.cells.forEach((cell) => {
+      if (identity === RunmeIdentity.UNSPECIFIED && cell.metadata?.['runme.dev/originalId']) {
```
sourishkrout, line 358: We should try to make `reconcileCellIdentity` go away instead of further expanding the quick fix band-aid that it was.

src/extension/serializer.ts
sourishkrout: One could argue that if `runme.dev/id` is not set the deserialization was requested without cell identity.

sourishkrout (changes requested):
Again, I think we should try to solve the issue at deserialization where it originates instead of "fixing" reconciliation which wasn't intended to be a permanent solution.

sourishkrout:
Closing as obsolete now that https://github.com/stateful/vscode-runme/pull/1449 is merged.

PR #1376, opened 2024-05-23
Add a warning to avoid conflicts across Stateful extensions

pastuxso:
Stateful Platform is a Superset of the Runme extension, so it is necessary to add a warning to avoid strange behaviour.


sourishkrout (approved):
✅  LGTM

PR #1378, opened 2024-05-24
Runner with v1 <> v2 toggle

sourishkrout:
Keep Runner v1 as default but allow v2 To allow for broader testing.

PS: C/I will fail until this PR's protos were published: https://github.com/stateful/runme/pull/589

### DIFF
--- a/src/extension/runner/index.ts
+++ b/src/extension/runner/index.ts
@@ -1,4 +1,3 @@
-import { DuplexStreamingCall } from '@protobuf-ts/runtime-rpc/build/types/duplex-streaming-call'
 import {
   type Pseudoterminal,
   type Event,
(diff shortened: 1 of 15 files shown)


src/extension/runner/index.ts
```diff
@@ -279,13 +282,15 @@ export default class GrpcRunner implements IRunner {
...
-    const { session } = await this.client.getSession(GetSessionRequest.create({ id })).response
+    const { session } = await this.client.getSession(GetSessionRequestImpl().create({ id }))
+      .response
 
     if (!session) {
       return undefined
     }
 
-    convertEnvList(session.envs)
+    const env = (session as any).envs ?? (session as any).env ?? []
```
pastuxso, line 292: It looks redundant, but after a review of other alternatives, I think the current one makes sense. :+1:

pastuxso (approved):
Running locally, everything works :ok_hand:. But please review my comments.

image

Issue #1374, opened 2024-05-22
Disabled `Preview Gist` button when `runme.app.sessionOutputs` is disabled


sourishkrout:
To avoid confusion because without Session Outputs, Gist won't work.

<img width="1947" alt="image" src="https://github.com/stateful/vscode-runme/assets/250527/a37ed943-e953-4eab-b544-e59335dcfb16">

As it stands, the `Preview Gist` button can't ever made work and not displaying is the solution.

PR #1388, opened 2024-05-28
Disable "Preview Gist" when session outputs are off

pastuxso:
Right now, the "Preview Gist" button appears even when session outputs are disabled, so to avoid confusion, this PR disables the button when the session outputs setting is off.

Closes: #1374


sourishkrout (approved):
LGTM ✅

PR #1390, opened 2024-05-29
Use kernel for auto media type detection

sourishkrout:
- [x] Use kernel detection for media type
- [x] Change `mimeType` description in annotations foldout
- [x] Remove Vercel legacy code
- [x] Consider using explicit `mimeType` as priority (not fallback) over auto-detection


pastuxso (approved):
LGTM! :+1:

PR #1413, opened 2024-06-07
Make the Gist listings on GH profile page more useful

sourishkrout:
Summary last, as in prefix with "summary-" which alphabetically gives it lower priority.

### DIFF
--- a/src/extension/commands/index.ts
+++ b/src/extension/commands/index.ts
@@ -391,14 +391,14 @@ export async function createGistCommand(e: NotebookUiEvent, context: ExtensionCo
         const createdGist = await createGist({
           isPublic: false,
           files: {
-            [sessionId]: {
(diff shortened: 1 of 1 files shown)


pastuxso (changes requested):
I think you should consider file names starting after the s. i.e. **t, u, w, etc.**

image

sourishkrout:
> I think you should consider file names starting after the s. i.e. **t, u, w, etc.**

I know, but I really think leaving it up to "alphabetical chance" is honest and if the summary shows up first accordingly. That's okay. We just don't want it to be unfairly prioritized, @pastuxso.

pastuxso:
@sourishkrout Thanks for clarifying.

pastuxso (approved):
LGTM!

PR #1415, opened 2024-06-08
Add git context

peraltafederico:
Introduces functionality to send Git context data to Platform by spawning Git commands.


sourishkrout (commented):
In don't love this because VS Code does have Git Information APIs (integral part of `vscode.git`). However, I think we can replace it later since this only runs for the platform.

@pastuxso can you do a functional review, please?

pastuxso (approved):
LGTM 👍

PR #1441, opened 2024-06-24
Run tests against latest vscode

sourishkrout:
We had issues when VS Code v1.90 was first released. Let's see if that's resolved.


pastuxso (approved):
LGTM 👍

PR #1443, opened 2024-06-25
Prevent notebook UX from being rendered unusable

sourishkrout:
Catch the exception and stop cell execution as unsuccessful to prevent "runaway" execution as shown here:

https://github.com/stateful/vscode-runme/assets/250527/6e3dad27-bd40-4140-bade-22d98a866fc4

At some future point, I'd like to do a more substantial refactor of how Cloud Renderers are "embedded" in bash. However, we need to collect more feedback from usage first.


pastuxso (approved):
LGTM 👍

PR #1451, opened 2024-06-29
Config NODE_OPTIONS

peraltafederico:
Config `NODE_OPTIONS` in `test` and `test:unit` so node can resolve ESM modules.


pastuxso (approved):
LGTM 👍

PR #1452, opened 2024-06-29
Use native xvfb

peraltafederico:
Use native xvfb instead of `coactions/setup-xvfb@v1` because it's failing for GH pipelines. See  https://github.com/coactions/setup-xvfb/issues/19

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -60,6 +60,9 @@ jobs:
         with:
           node-version: 18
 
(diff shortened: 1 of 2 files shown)


pastuxso:
@peraltafederico I think it's unnecessary. Unfortunately, this Github Action lacks graceful termination, causing a false negative.

pastuxso (approved):
LGTM

PR #1454, opened 2024-06-29
Implements Stateful Platform Authentication on Startup 

pastuxso:
If no Platform Authentication exists on startup, a new stateful session will be created.


sourishkrout (approved):
👍 LGTM

PR #1465, opened 2024-07-06
Initial GCE Instance view

pastuxso:
Adds initial rendering for CGE instances. Some information is unavailable in the first API request, but it's sufficient to show relevant information. I only needed to make a second API request to complete the disk information.


sourishkrout (approved):
:+1: LGTM

PR #1467, opened 2024-07-09
More descriptive marketplace description

admc:
Might help to more explicitly explain here what these things are, sort of define DevOps Notebook in a way.


sourishkrout (approved):
👍 LGTM

PR #1473, opened 2024-07-10
Runme in Stateful package description

admc:
I think it's importat to describe a relationship with runme


sourishkrout (approved):
👍  LGTM

PR #1474, opened 2024-07-10
Start moving away from runme cloud domain/endpoint resolution

sourishkrout:
This PR moves away from any panels served out of the legacy Runme cloud.

Let's plan for a follow-up effort to remove all non-platform panels/GQL since the backend is considered End Of Life.


pastuxso (approved):
LGTM 👍

PR #1476, opened 2024-07-11
Stateful extension marketplace README.md

admc:
The market place is a good marketing opportunity, we should have a readme that talks about Stateful as an integrated solution.

I took a first stab at it, if you folks can wire up the overwrite for publishing that would be awesome.

@sourishkrout wdyt?

image


sourishkrout (approved):
Let's try 👍

PR #1479, opened 2024-07-12
Improvements to `Env Store` (experimental) panel

sourishkrout:
- Better tooltip visibility triggered by hovering over any value, not just the leading icon
- Semantic sorting by error type, which brings invalid ENV vars to the top
- Cosmetic tweaks


src/client/components/env/store.ts
```diff
@@ -99,38 +113,63 @@ export default class Table extends LitElement {
...
+          .tooltipText="${html`<div class="flex">
+            ${row.errors.map((error) => html`<span>${icon}${error.message}</span>`)}
+          </div>`}"
+          .value="${html`${icon} ${format()}`}"
+        ></tooltip-text>
+      </div>`
+    }
+
+    if (field === 'spec') {
+      // if (!['Opaque', 'Plain', 'Secret', 'Password'].includes(row.spec)) {
```
pastuxso, line 161: Is it intentionally commented?

pastuxso (approved):
LGTM 🚀

src/client/components/env/store.ts
sourishkrout: yep, it's incomplete

PR #1487, opened 2024-07-18
Hide preview gist button

peraltafederico:
Hide the Preview Gist button when running the extension with Platform authentication.


pastuxso:
✅ LGTMT

PR #1488, opened 2024-07-18
Platform auth secrets per stage

sourishkrout:
I believe the reason why we see "false positive" errors when switching stages is that auth tokens aren't valid across stages (staging, prod, etc). The solution is fairly simple by stashing the token into the secrets store under keys that are unique to the stages.


pastuxso (approved):
LGTM 👍

PR #1489, opened 2024-07-19
Simplify and remove deletion of cell ID

sourishkrout:
The deletion of the cell ID causes downstream issues. It also needs to be clarified why it was required in the first place.

I removed the logic to test it for a few days in pre-release.


pastuxso (approved):
LGTM 👍

PR #1496, opened 2024-07-26
Setting to let users decide if masked or unmasked should be default

sourishkrout:
Solves https://github.com/stateful/runme/issues/639.


sourishkrout:
@jlewi lemme know if you want to do the review. If not, @pastuxso can take it.

jlewi:
LGTM.

Thanks for the quick fix.

sourishkrout:
Included in Runme's latest pre-release effectively now. It should be stable in the next 24-48 hours.

PR #1509, opened 2024-07-31
Fork/rollup terminal

sourishkrout:
First stab at unlocking Runme Terminal features. Only for runner V2.

Requires kernel binary with https://github.com/stateful/runme/pull/638.


pastuxso (approved):
LGTM 🚀

PR #1515, opened 2024-08-01
Adds support to render Regional GKE Clusters

### LINKED ISSUE #1510: GKE Renderer: Invalid value for zone: us-west1
I tried the GKE renderer. I pasted the link into a code cell

```
https://console.cloud.google.com/kubernetes/list/overview?hl=en&project=contr-jeremylewi
```

I got the error

```
Failed to get Google Cloud resource data, reason: Invalid value for field 'zone': 'us-west1'. Unknown zone.
```

`us-west1` is a region not a zone so that would nominally explain the error. I have a regional cluster in us-west1 but not a zonal cluster.

pastuxso:
The bug is occurring because the request to retrieve the machine type requires the zone, but we were passing the region when the cluster is actually regional.

In the case of a regional cluster, multiple zones are listed in `cluster.locations`. Therefore, it is now necessary to display the count of nodes, total CPU, and total memory requesting the machine type for each zone.

Closes #1510

### DIFF
--- a/src/extension/executors/gcp/gke/clusters.ts
+++ b/src/extension/executors/gcp/gke/clusters.ts
@@ -31,34 +31,52 @@ export async function getClusters(
   const projectId = useDefaultProject ? await clusterManagement.getProjectId() : project
 
   const [response] = await clusterManagement.listClusters({
-    parent: `projects/${projectId}/locations/-`,
+    projectId: projectId,
+    zone: '-',
   })
+
   if (!response.clusters) {
     return
   }
+
   const clusters: GcpGkeCluster[] = []
 
   for await (const cluster of response.clusters) {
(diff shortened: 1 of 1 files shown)


jlewi:
@pastuxso Thanks for the quick fix. Could you explain what the bug was and how this fixes it?

sourishkrout (approved):
✅ works. if you could answer @jlewi's question, please @pastuxso.

pastuxso:
@jlewi 

> @pastuxso Thanks for the quick fix. Could you explain what the bug was and how this fixes it?

Sure, I've updated the pull request description to provide more details about the changes. Feel free to ask me if you have any questions.

jlewi:
Great thank you.
LGTM

PR #1517, opened 2024-08-02
Add open cell output button

peraltafederico:
Add an "Open" button when a cell output is saved. Clicking that button redirects the user to the Platform.


pastuxso (approved):
LGTM 👍

PR #1532, opened 2024-08-08
Fix git dir

peraltafederico:
Use the forwarded `editor` instead of `window.activeTextEditor` to get the current file path of a git folder.


sourishkrout (approved):
✅ LGTM. Always better to not rely on "global" state.

PR #1536, opened 2024-08-13
Make ghost cells executable by preserving the runme.dev/id metadata

jlewi:
Fix jlewi/foyle#188

The reason newly added ghost cells aren't executable is because they they lack a runme.dev/id metadata. This field gets set when cells are inserted but it was being removed by the code that updates the cell to set ghost to false. This updateMetadata calls was replacing the entire metadata with a new dictionary and not just updating it. So to fix it we need to pass in a dictionary that contains existing metadata that we want to preserve.

### DIFF
--- a/src/extension/ai/ghost.ts
+++ b/src/extension/ai/ghost.ts
@@ -335,8 +335,9 @@ export function handleOnDidChangeActiveTextEditor(editor: vscode.TextEditor | un
   if (!isGhostCell(cell)) {
     return
   }
-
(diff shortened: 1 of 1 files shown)


jlewi:
@sourishkrout PTAL

jlewi:
@sourishkrout friendly nudge

sourishkrout:
> @sourishkrout friendly nudge

👌. I will take a look over the weekend.

sourishkrout (approved):
👍 LGTM

Issue #1393, opened 2024-05-30
Detect and warn when unsupported shells are being used (e.g. fish)


sourishkrout:
Runme should be able to detect when an unsupported shell is being used via `$SHELL` and `$0` and could warn accordingly.

The current error message looks something like this which isn't particularly helpful.

```error: Internal failure executing runner: open /var/folders/hn/3tbqz8pd49q9yl2ns43qx6fh0000gp/T/512117330/.env_start: no such file or directory```

Let's detect unsupported shells (e.g., fish, PowerShell).

Plan to show warning message with link to documentation for unsupported shells.

Provide button to open Documentation URL to be configurable for potential Stateful-specific overrides.

sourishkrout:
Let's add a "Don't ask again" option to the notification, similar to https://github.com/stateful/vscode-runme/blob/main/src/extension/messaging.ts#L100-L102, @pastuxso, please. I think we want that before we can turn on the warning by default.

sourishkrout:
@pastuxso let's flip the experiment to be default `true` to expose it publicly.

PR #1550, opened 2024-08-16
Warn about unsupported shells

pastuxso:
This PR introduces changes to make a best effort to determine if the shell is supported.

Progress: #1393

### DIFF
--- a/src/extension/executors/aws.ts
+++ b/src/extension/executors/aws.ts
@@ -47,8 +47,7 @@ export const aws: IKernelExecutor = async (executor) => {
         break
     }
 
-    const result = await kernel.runProgram(programOptions)
-    const profile = result.code === 0 ? result.output : 'default'
+    const profile = await kernel.runProgram(programOptions)
     credentials = fromIni({ profile })
 
     switch (awsResolver.view) {
--- a/src/extension/executors/aws.ts
+++ b/src/extension/executors/aws.ts
@@ -47,7 +47,13 @@ export const aws: IKernelExecutor = async (executor) => {
         break
     }
 
-    const profile = await kernel.runProgram(programOptions)
+    let profile = ''
+    const result = await kernel.runProgram(programOptions)
+
+    if (result) {
+      profile = result
+    }
+
     credentials = fromIni({ profile })
 
     switch (awsResolver.view) {
--- a/src/extension/executors/aws.ts
+++ b/src/extension/executors/aws.ts
@@ -14,7 +14,7 @@ import { resolveProgramOptionsScript } from './runner'
 import { IKernelExecutor } from '.'
 
 export const aws: IKernelExecutor = async (executor) => {
-  const { cellText, exec, runner, runnerEnv, doc, outputs, context } = executor
+  const { cellText, exec, runner, runnerEnv, doc, outputs, kernel } = executor
 
   const annotations = getAnnotations(exec.cell)
 
@@ -41,67 +41,14 @@ export const aws: IKernelExecutor = async (executor) => {
       cellId,
     })
 
-    // todo(sebastian): move down into kernel?
     switch (programOptions.exec?.type) {
       case 'script':
-        {
-          programOptions.exec.script = 'echo $AWS_PROFILE'
-        }
+        programOptions.exec.script = 'echo $AWS_PROFILE'
         break
     }
 
-    const program = await runner.createProgramSession(programOptions)
-    context.subscriptions.push(program)
-
-    let execRes: string | undefined
-    const onData = (data: string | Uint8Array) => {
-      if (execRes === undefined) {
-        execRes = ''
-      }
-      execRes += data.toString()
-    }
-
-    program.onDidWrite(onData)
-    program.onDidErr(onData)
-    program.run()
-
-    const success = await new Promise<boolean>((resolve, reject) => {
-      program.onDidClose(async (code) => {
-        if (code !== 0) {
-          return resolve(false)
-        }
-        return resolve(true)
-      })
-
-      program.onInternalErr((e) => {
-        reject(e)
-      })
-
-      const exitReason = program.hasExited()
-
-      // unexpected early return, likely an error
-      if (exitReason) {
-        switch (exitReason.type) {
-          case 'error':
-            {
-              reject(exitReason.error)
-            }
-            break
-
-          case 'exit':
-            {
-              resolve(exitReason.code === 0)
-            }
-            break
-
-          default: {
-            resolve(false)
-          }
-        }
-      }
-    })
-
-    const profile = success ? execRes?.trim() : 'default'
+    const result = await kernel.runProgram(programOptions)
+    const profile = result.code === 0 ? result.output : 'default'
     credentials = fromIni({ profile })
 
     switch (awsResolver.view) {
--- a/src/extension/extension.ts
+++ b/src/extension/extension.ts
@@ -80,6 +80,7 @@ import { NotebookPanel as EnvStorePanel } from './panels/notebook'
 import { NotebookCellStatusBarProvider } from './provider/cellStatusBar/notebook'
 import { SessionOutputCellStatusBarProvider } from './provider/cellStatusBar/sessionOutput'
 import * as manager from './ai/manager'
+import getLogger from './logger'
 
 export class RunmeExtension {
   protected serializer?: SerializerBase
@@ -370,19 +371,21 @@ export class RunmeExtension {
         }
       }
 
+      const logger = getLogger('runme.beta.shellWarning')
+
       kernel
         .runProgram('echo $SHELL')
-        .then(({ output }) => {
+        .then((output) => {
           const supportedShells = ['bash', 'zsh']
           const isSupported = supportedShells.some((sh) => output?.includes(sh))
-          console.log(`[runme.beta.shellWarning] Shell: ${output}`)
+          logger.info(`[runme.beta.shellWarning] Shell: ${output}`)
 
           if (!isSupported) {
             showUnsupportedShellMessage()
           }
         })
         .catch((e) => {
-          console.log(e)
+          logger.error(e)
           showUnsupportedShellMessage()
         })
     }
--- a/src/extension/extension.ts
+++ b/src/extension/extension.ts
@@ -376,6 +376,10 @@ export class RunmeExtension {
       kernel
         .runProgram('echo $SHELL')
         .then((output) => {
+          if (output === false) {
+            return
+          }
+
           const supportedShells = ['bash', 'zsh']
           const isSupported = supportedShells.some((sh) => output?.includes(sh))
           logger.info(`Shell: ${output}`)
--- a/src/extension/extension.ts
+++ b/src/extension/extension.ts
@@ -15,7 +15,6 @@ import Channel from 'tangle/webviews'
 
 import { NotebookUiEvent, Serializer, SyncSchema } from '../types'
 import {
-  getBetaFlag,
   getDocsUrl,
   getForceNewWindowConfig,
   getRunmeAppUrl,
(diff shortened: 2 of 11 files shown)


src/extension/executors/aws.ts
```diff
@@ -41,67 +41,14 @@ export const aws: IKernelExecutor = async (executor) => {
...
     switch (programOptions.exec?.type) {
       case 'script':
-        {
-          programOptions.exec.script = 'echo $AWS_PROFILE'
-        }
+        programOptions.exec.script = 'echo $AWS_PROFILE'
         break
     }
 
-    const program = await runner.createProgramSession(programOptions)
```
pastuxso, line 53: I took advantage to avoid code duplication.

src/extension/kernel.ts
```diff
@@ -1119,4 +1119,79 @@ export class Kernel implements Disposable {
...
+          return resolve({ code, output: execRes?.trim() })
+        })
+
+        programSession.onInternalErr((e) => {
+          reject(e)
+        })
+
+        const exitReason = programSession.hasExited()
+
+        if (exitReason) {
```
sourishkrout, line 1175: When `exitReason` is null/undefined, which isn't impossible, the promise will dangle (never resolve or reject). The fix is likely to switch against `exitReason?.type` and make the default case error.

src/extension/extension.ts
```diff
@@ -353,6 +356,27 @@ export class RunmeExtension {
...
+            if (answer === showMore) {
+              const dashboardUri = getFaqUrl('supported-shells')
+              const uri = Uri.parse(dashboardUri)
+              return env.openExternal(uri)
+            }
+          })
+        }
+      })
+      .catch((_e) => {
+        getLogger().warn('An error occurred while verifying the supported shell')
```
sourishkrout, line 377: If we can 100% make sure you have a compatible shell, shouldn't that be a reason to error/warn about the shell?

src/extension/extension.ts
```diff
@@ -353,6 +356,27 @@ export class RunmeExtension {
...
 
+    kernel
+      .runProgram('echo $0')
+      .then(({ output }) => {
+        const supportedShells = ['zsh', 'bash']
+        const isSupported = supportedShells.some((sh) => output?.includes(sh))
+
+        if (!isSupported) {
+          const showMore = 'Show more'
+          return window.showWarningMessage('Unsupported shell', showMore).then((answer) => {
```
sourishkrout, line 367: Is it an error or a warning? We can recover when somebody's running `pwsh` or `cmd.exe` or `fish`. Smells like error to me.

src/extension/kernel.ts
```diff
@@ -1119,4 +1119,79 @@ export class Kernel implements Disposable {
...
+
+    if (!this.runner) {
+      return Promise.reject(new Error('No runner available'))
+    }
+
+    if (typeof program === 'object') {
+      programOptions = program
+    } else if (typeof program === 'string') {
+      programOptions = {
+        programName: '',
```
sourishkrout, line 1134: I don't think this can be empty. We should probably use `bash` (if we can rely on $PATH) or `/bin/bash` and perhaps check the default shell `$SHELL` instead of `$0` (what's shell is currently running)?

src/extension/executors/aws.ts
sourishkrout: Good idea to DRY. This will do for the simple cases.

src/utils/configuration.ts
```diff
@@ -432,6 +434,11 @@ const isPlatformAuthEnabled = (): boolean => {
   return getCloudConfigurationValue('platformAuth', false)
 }
 
+const getFaqUrl = (hash: string): string => {
```
sourishkrout, line 437: I don't love this. I'd suggest using a purpose-specific setting and leveraging a default URL that uses the destination redirection pattern (as used in the foldout links):

https://github.com/stateful/docs.runme.dev/blob/main/vercel.json#L3-L62

That way it's easy to change the final docs destination and move it around even for older packaged versions.

sourishkrout (changes requested):
Please see comments.

src/extension/kernel.ts
sourishkrout: Nvm. I just checked the "reference impl," and this code checks for premature termination of the program, so it's not expected always to be non-undefined. In fact, if it's defined it's almost always an error.

src/extension/kernel.ts
```diff
@@ -1119,4 +1119,79 @@ export class Kernel implements Disposable {
...
+    let programOptions: RunProgramOptions
+
+    if (!this.runner) {
+      return Promise.reject(new Error('No runner available'))
+    }
+
+    if (typeof program === 'object') {
+      programOptions = program
+    } else if (typeof program === 'string') {
+      programOptions = {
```
sourishkrout, line 1133: This won't work without `commandMode: CommandModeEnum().INLINE_SHELL,`

src/extension/kernel.ts
```diff
@@ -1119,4 +1119,79 @@ export class Kernel implements Disposable {
...
+    const onData = (data: string | Uint8Array) => {
+      if (execRes === undefined) {
+        execRes = ''
+      }
+      execRes += data.toString()
+    }
+
+    programSession.onDidWrite(onData)
+    programSession.onDidErr(onData)
+    await programSession.run()
```
sourishkrout, line 1161: You can't await this before "wiring up" the promise. It'll block here indefinitely. That's likely why the callee can't currently be awaited. It just hangs.

src/extension/extension.ts
```diff
@@ -354,6 +357,34 @@ export class RunmeExtension {
...
+        if (answer === showMore) {
+          const url = getDocsUrl('/r/extension/supported-shells')
+          env.openExternal(Uri.parse(url))
+        }
+      }
+
+      kernel
+        .runProgram('echo $SHELL')
+        .then(({ output }) => {
+          const supportedShells = ['bash']
```
sourishkrout, line 375: @pastuxso no more `zsh`?

src/utils/configuration.ts
```diff
@@ -432,6 +437,21 @@ const isPlatformAuthEnabled = (): boolean => {
   return getCloudConfigurationValue('platformAuth', false)
 }
 
+const getDocsUrl = (path: string): string => {
+  const baseUrl = getCloudConfigurationValue('docsUrl', DEFAULT_DOCS_URL)
+  return `${baseUrl}#${path}`
+}
+
+const getBetaFlag = <T>(configName: keyof typeof configurationSchema.beta, defaultValue: T) => {
```
sourishkrout, line 445: @pastuxso why not reuse the experiments?

https://github.com/stateful/vscode-runme/blob/cris/unsupported-shell-warn/src/extension/kernel.ts#L138-L143

src/utils/configuration.ts
```diff
@@ -90,6 +91,10 @@ const configurationSchema = {
     maskOutputs: z.boolean().default(true),
     loginPrompt: z.boolean().default(true),
     platformAuth: z.boolean().default(false),
+    docsUrl: z.string().default(DEFAULT_DOCS_URL),
+  },
+  beta: {
+    shellWarning: z.boolean().default(true),
```
sourishkrout, line 97: @pastuxso default `true`? again why not use experiments instead of `beta`

sourishkrout (changes requested):
See comments.

src/extension/extension.ts
```diff
@@ -354,6 +357,34 @@ export class RunmeExtension {
...
+          const isSupported = supportedShells.some((sh) => output?.includes(sh))
+          console.log(`[runme.beta.shellWarning] Shell: ${output}`)
```
sourishkrout, line 377: why not use the logger for this?

src/extension/kernel.ts
```diff
@@ -1102,4 +1102,81 @@ export class Kernel implements Disposable {
   public getPlainCache(cacheId: string): Promise<Uint8Array> | undefined {
     return this.serializer?.getPlainCache(cacheId)
   }
+
+  runProgram(program?: RunProgramOptions | string) {
+    return new Promise<{ exitReason?: RunnerExitReason; code?: number | void; output?: string }>(
```
sourishkrout, line 1107: @pastuxso this is still not right. i'm going to send a PR fixing `resouce.ts` that you can model it after.

sourishkrout: https://github.com/stateful/vscode-runme/pull/1582

sourishkrout (changes requested):

src/utils/configuration.ts
pastuxso: 🙌  Well spotted! It's crucial to ensure that I'm testing beyond just the happy path.

src/extension/extension.ts
pastuxso: Thank you for catching that. I thought I had added it back, but it seems I missed it. I’ll correct it right away. I appreciate your attention to detail.

src/extension/kernel.ts
```diff
@@ -141,6 +141,7 @@ export class Kernel implements Disposable {
     this.#experiments.set('escalationButton', config.get<boolean>('escalationButton', false))
     this.#experiments.set('smartEnvStore', config.get<boolean>('smartEnvStore', false))
     this.#experiments.set('aiLogs', config.get<boolean>('aiLogs', false))
+    this.#experiments.set('shellWarning', config.get<boolean>('shellWarning', false))
```
sourishkrout, line 144: I don't see this experiment defined in `package.json`.

src/extension/kernel.ts
```diff
@@ -1102,4 +1103,86 @@ export class Kernel implements Disposable {
   public getPlainCache(cacheId: string): Promise<Uint8Array> | undefined {
     return this.serializer?.getPlainCache(cacheId)
   }
+
+  async runProgram(program?: RunProgramOptions | string) {
+    let programOptions: RunProgramOptions
+
+    if (!this.runner) {
+      throw new Error('No runner available')
```
sourishkrout, line 1111: Don't think this should be "fatal" and throw. We can recover by logging a warning here and return early. If the `runme.experiments.grpcRunner` is disabled this will make Runme hard stop for no reason.

src/types.ts
```diff
@@ -647,6 +643,12 @@ export interface OutputItemsPayload {
   id: string
 }
 
+export interface AnnotationsPayload {
```
sourishkrout, line 646: Perhaps moving this up to be more colocated with where's being used. Seems odd so far down since annotations are so integral to Runme.

sourishkrout (changes requested):
Getting really close. More comments.

src/extension/kernel.ts
pastuxso: @sourishkrout I'm aware of that. I think adding the respective entry in the package.json is unnecessary for this kind of temporary flag. But, if you feel it's necessary I'll add it.

src/types.ts
pastuxso: Good point. I moved it up close to the previous one. Let me move this to a better place.

src/extension/kernel.ts
pastuxso: Oh, I see. It’s a leftover from the previous approach. In this case, I think we can return false instead

src/extension/kernel.ts
sourishkrout: let's add it. i think it could come handy to enable user to turn it off if they experience problems. more self-explanatory if it's included in `package.json`.

src/extension/kernel.ts
sourishkrout: Whatever makes it noop. If the runner is disabled, we should never display a notification.

src/extension/kernel.ts
pastuxso: @sourishkrout added ✅

sourishkrout:
image

@pastuxso the early return is not being handled properly. in the case where `grpcRunner` is not `true` we don't want the error message aka we do want noop.

sourishkrout (changes requested):
See comment.

sourishkrout (approved):
✅ LGTM

PR #1559, opened 2024-08-19
Use the connect protocol to generate unary AI completions

jlewi:
* This PR switches from using gRPC and the timostamm client to using the connect client for unary completions. This makes the code consistent with what we are doing for the streaming completions with ghost cells.

* The service definition is also published is part of Foyle's protos in the buf schema registry https://buf.build/jlewi/foyle/commits/commit/b69cd71876b144ca900524437d722a84 so we can remove the AI service defined in the runme buf package (https://github.com/stateful/runme/tree/main/pkg/api/proto/runme/ai) after this change is released.

* The benefit of this change for users is that we can have a single configuration option AI Base URL for configuring the AIService. This will be used for all AI calls (unary or streaming). We can then remove the option FoyleAddress.

* This will also allow us to simplify the Foyle server since we will be able to stop running gRPC and gRPC gateway. see jlewi/foyle#173


jlewi:
Need to merge #1554 and then rebase this PR.
We should be able to drop all the changes prior to commit [7c3924a](https://github.com/stateful/vscode-runme/pull/1559/commits/7c3924a625757ebf81aeac2aa7f4a5b178cefe1e)

jlewi:
@sourishkrout this is ready for review.

sourishkrout:
> @sourishkrout this is ready for review.

Sounds good. Feel free to add me as one of the reviewers. A Slack bot reminds me twice a day about pending reviews.

tests/extension/extension.test.ts
```diff
@@ -74,6 +74,8 @@ vi.mock('../../src/extension/grpc/runner/v1', () => ({}))
 test('initializes all providers', async () => {
   const configValues = {
     binaryPath: 'bin',
+    // This is needed for the AIManager to initialize.
+    'runme.aiBaseURL': 'http://localhost:8080/api',
```
sourishkrout, line 78: I'd suggest using a different port, `8080`, which is over-used. Fine inside container images, but running locally, there's a high likelihood another process bound the port locally.

sourishkrout (approved):
✅ LGTM

tests/extension/extension.test.ts
jlewi: Good Idea. I changed the default to 8877.
WDYT?

sourishkrout: Good choice, @jlewi!

PR #1578, opened 2024-08-21
Exclude GitHub Action e2e Tests for Dependabot PRs

pastuxso:
Excludes the end-to-end test that always fails due to permissions. Additionally, the auto-merge was disabled because we are doing it manually.

Having a CI passing ✅  helps to diagnose potential breaking changes in advance.


sourishkrout (approved):
✅ LGTM

PR #1586, opened 2024-08-26
Rename runner v2alpha1 to v2

sourishkrout:
Tracking the runner's version promotion from `v2alpha1` to `v2` in the upstream protos.


sourishkrout:
Please review, @pastuxso. This is merely a version name change (promotion) from `v2alpha1` to `v2`.

I am planning to release with `runme` kernel binary 3.7. However, older versions should continue to work because the protocol has stayed the same on the binary level.

pastuxso:
LGTM 🚀

pastuxso (approved):

PR #1589, opened 2024-08-28
Use the Foyle API to report Log Events

jlewi:
* For motivation and description see: https://github.com/jlewi/foyle/issues/211

* This PR allows us to log all executions along with some context for the execution. This will allow us to improve learning. Right now we don't learn unless a user accepts a suggestion and edits it. This will allow us to learn even if the user didn't use a suggested cell.

* This PR introduces a SessionManager for the AI. A session is initiated each time the user switches focus to a different cell. The session corresponds to all the activity related to generating completions as the user edits that cell. A session is associated with a context (basically the notebook) as well as events (cell executions, suggestion acceptances). We log the start and end of the session.

* Logging the start and end of the session should simplify log processing on the backend because now we know when a session is closed and there can be no more events associated with the session. Notably, we want to be able to determine which suggestions were rejected rather than accepted. Once a session is closed we know the suggestion was either accepted (there will be an accepted event) or the suggestion was not accepted.


jlewi:
@sourishkrout PTAL when you have a second.

J

sourishkrout (approved):
✅ LGTM. The only nit I have is perhaps adding a test to guard against accidental removal/change of the event reporter.

`tests/extension/kernel.test.ts` should have sufficiently mocked examples to test for it. I'm happy to push a test on top if you want me to write one. Lemme know.

sourishkrout:
@jlewi, will this make the server side AI user logs entirely obsolete? I'm asking because I have a todo to port them from `v1` to the `v2` runner. Will that be moot now?

src/extension/kernel.ts
```diff
@@ -691,6 +692,7 @@ export class Kernel implements Disposable {
     }
 
     TelemetryReporter.sendTelemetryEvent('cell.startExecute')
+    getEventReporter().reportExecution(cell)
```
sourishkrout, line 695: the reporter call is async, we should probably await it here

src/extension/ai/events.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    contextCells.push(cell)
+    const cells = converters.vsCellsToESProtos(contextCells)
+    const event = new LogEvent()
+    event.selectedId = cell.metadata?.[RUNME_CELL_ID]
+    event.selectedIndex = cell.index
+    event.type = LogEventType.EXECUTE
+    event.cells = cells
+    event.contextId = SessionManager.getManager().getID()
+    this.reportEvents([event])
```
sourishkrout, line 61: This should be returned to pass up the promise, expects `Promise<void>`.

src/extension/ai/events.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  async reportEvents(events: LogEvent[]) {
+    const req = new LogEventsRequest()
+    req.events = events
+    for (const event of events) {
+      if (event.eventId === '') {
+        event.eventId = ulid()
+      }
+    }
+    await this.client.logEvents(req).catch((e) => {
```
sourishkrout, line 72: return instead of await here otherwise the call stack won't account for asynchronicity

sourishkrout (changes requested):
Actually I did find some issues with async/await.

src/extension/ai/events.ts
sourishkrout: an `await` should to it too

src/extension/ai/events.ts
sourishkrout: actually if the return value is irrelevant the `await` will do

jlewi:
> @jlewi, will this make the server side AI user logs entirely obsolete?

Yes. I will send you a follow on PR to remove the EnableAILogs option in the frontend.

I was going to ask you if you wanted me to send you a PR to revert/cleanup much of stateful/runme#585

src/extension/kernel.ts
jlewi: Would that block execution below? My thinking was that we want LogEvent Reporting to be out of band and not block critical processing. If I was doing this in go I'd fire off a go routine. Is calling an async function and not awaiting it similar to doing `go somefunc`?

sourishkrout:
> Yes. I will send you a follow on PR to remove the EnableAILogs option in the frontend.
> 
> I was going to ask you if you wanted me to send you a PR to revert/cleanup much of [stateful/runme#585](https://github.com/stateful/runme/pull/585)

Yes, on the cleanup PR, please.

Btw, I just pushed https://github.com/stateful/vscode-runme/pull/1602 to add testing, which I needed for the review anyway.

src/extension/ai/events.ts
jlewi: Added it.

src/extension/ai/events.ts
jlewi: I think if we use a return here it will change the return type to Promise<LogEventsResponse>
I think Promise<void> is the better return type because I don't think the caller of the reporter should be trying to process any results of the reporter.

jlewi:
> Btw, I just pushed https://github.com/stateful/vscode-runme/pull/1602 to add testing, which I needed for the review anyway

Thanks!

jlewi:
stateful/runme#661 has the gRPC changes

src/extension/kernel.ts
jlewi, line 695: @sourishkrout I'm replying to your comment 
https://github.com/stateful/vscode-runme/pull/1602#pullrequestreview-2273121720
about adding an await here;  on the original PR.

> Yes. I incorrectly assumed the noop event reporter would be unless the "AI experiment" is turned on.

That was the intent. Is there a bug here that I need to fix? This is how I intended it to work.

The reporter should be the null op reporter by default 
https://github.com/stateful/vscode-runme/pull/1589/files#diff-f0a5c022488c3922874e791475fce22933ea551ffbc4a37c414de2831f8af593R89

and then when the extension loads if the AutoCellExperiment is turned on it get changed to the actual event reproter.
https://github.com/stateful/vscode-runme/pull/1589/files#diff-d1df40b876292a3e89a6d7d6d6e7496db98fa7ad03028b3b0f78bc90487749bdR30

One potential issue I see is that if you disable AIAutoCell the reporter won't be switched back to the null op reporter until the extension is reloaded. I assume that can be triggered by doing a reload window?
That's not great but I didn't know how to subscribe to notifications when options changed and I figured its probably good enough for now. 

> I can rewrite this reasonably quickly at a later point. Wdyt @jlewi?
If your good with the current implementation then I'm happy.

For users trying out AI and Foyle I think this fixes a big UX issue. In particular, this will remove the need to explicitly add the location of the RunMe logs to the Foyle config. This also means we can turn on learning by default in Foyle (AiAutoCell would still be disabled by default in RunMe). So I see this is as a strict improvement.

The risk is that reporting cell executions causes a degraded experience. I've tried to mitigate that with the NullOp reporter. If users don't enable AI via the AutoCell option then no additional logic (modulo the actuall noop invocation) should be invoked.

sourishkrout: Let's add a `// todo(sebastian): rewrite to use non-blocking impl` on top of the `reportExecution ` call and merge for now. 👍 

K, my bad. I turned off AI logs not `aiAutoCell`. In any case, as long as it's not default we're good to go.

Re, settings requiring a hard reload: that's consistent with all other settings and not something to worry about in this context. Ghost cell completing might be something we want to move into the notebook toolbar and have a setting for the default state—however, one thing at a time.

sourishkrout: Here we go: https://github.com/stateful/vscode-runme/pull/1602/commits/bc2b0ac881d9008eb8c671a1645a37a758b4a083

jlewi: @sourishkrout Thanks. I patched it. Your patch has an `await in it`. Did you mean to include the await?

src/extension/kernel.ts
```diff
@@ -691,6 +692,8 @@ export class Kernel implements Disposable {
     }
 
     TelemetryReporter.sendTelemetryEvent('cell.startExecute')
+    // todo(sebastian): rewrite to use non-blocking impl
+    await getEventReporter().reportExecution(cell)
```
jlewi, line 696: @sourishkrout The await here came from your patch. Did you mean to include an await and make it blocking? If we call await here then I assume we block until the event is full reported? Whereas I thought if we don't use await here then reportExecution would perform the RPC asynchronously and not block actual execution?

jlewi: Thanks for the explanation!

jlewi: @sourishkrout Is this good to merge?

sourishkrout: Yep, approving the PR now.

sourishkrout (approved):
✅ LGTM. Will take care of a non-blocking event reporter at a later time.

PR #1590, opened 2024-08-28
Bug fix: AIManager doesn't properly get the configuration for baseURL

jlewi:
* We weren't actually reading the baseURL from the config because the key should be "aiBaseURL" not "runme.aiBaseURL"

* I tried to add logging to ensure we know what the actual value of the endpoint is but that seemed to prevent the extension from loading.

### DIFF
--- a/src/extension/ai/manager.ts
+++ b/src/extension/ai/manager.ts
@@ -17,12 +17,11 @@ export class AIManager {
   client: PromiseClient<typeof AIService>
   completionGenerator: generate.CompletionGenerator
   constructor() {
-    const config = vscode.workspace.getConfiguration('runme.experiments')
-    const autoComplete = config.get<boolean>('aiAutoCell', false)
-
-    this.client = this.createAIClient()
     this.log = getLogger('AIManager')
     this.log.info('AI: Initializing AI Manager')
+    const config = vscode.workspace.getConfiguration('runme.experiments')
+    const autoComplete = config.get<boolean>('aiAutoCell', false)
+    this.client = this.createAIClient()
     this.completionGenerator = new generate.CompletionGenerator(this.client)
     if (autoComplete) {
       this.registerGhostCellEvents()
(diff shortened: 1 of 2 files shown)


src/extension/ai/manager.ts
```diff
@@ -22,16 +22,28 @@ export class AIManager {
...
+    const baseURL = config.get<string>('aiBaseURL', 'http://localhost:8877/api')
+    // TODO(jeremy): The log.info statements prevents the extension from loading at least in debug
+    // mode. My suspicion
+    // is that createAIClient is invoked from the constructor which is invoked at extension
+    // loading time. In logger.ts there is a call to outputChannel = window.createOutputChannel('Runme')
+    // to setup the outputChannel. I wonder if that creates a race condition.
+    // The problem seemed to happen if I call console.log as well. It seemed somewhat intermittent
+    // Which would be consistent with a race condition. I wonder if the problem is specific to
+    // running in debug mode?
+    // this.log.info('AI: Using AI service at: ' + baseURL)
```
jlewi, line 44: @sourishkrout Do you have any  idea why these lines would prevent the extension from loading in debug mode in vscode?

sourishkrout: @jlewi I believe there's no `this` in this context because it's a standalone function based on its definition:

```
  createAIClient = (): PromiseClient<typeof AIService> => {
```

It might result in a static method. It's unclear without doing more research. Did I mean to make this a class member, i.e. what's below?

```
  createAIClient(): PromiseClient<typeof AIService> {
```

sourishkrout: Also, further up in the constructor, `createAIClient` is being called before the logger is initialized.

sourishkrout (approved):
Please see my comments @jlewi. Otherwise ✅.

src/extension/ai/manager.ts
jlewi: Of course its something obvious like that. Thanks for the help!

PR #1600, opened 2024-08-29
Marketplace naming update

admc:
Update descriptions to own the Runme Notebooks term a little more clearly


sourishkrout (approved):
👍 LGTM

PR #1601, opened 2024-08-30
Change versioning-strategy to increase

pastuxso:
This change is necessary to let Dependabot update the package.json too.


sourishkrout (approved):
✅ LGTM

PR #1602, opened 2024-08-30
Add tests to AI event logging

sourishkrout:
See diff, please, @jlewi.

### DIFF
--- a/src/extension/kernel.ts
+++ b/src/extension/kernel.ts
@@ -692,7 +692,7 @@ export class Kernel implements Disposable {
     }
 
     TelemetryReporter.sendTelemetryEvent('cell.startExecute')
-    getEventReporter().reportExecution(cell)
+    await getEventReporter().reportExecution(cell)
     runmeExec.start(Date.now())
 
     const annotations = getAnnotations(cell)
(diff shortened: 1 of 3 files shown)


src/extension/kernel.ts
```diff
@@ -692,7 +692,7 @@ export class Kernel implements Disposable {
     }
 
     TelemetryReporter.sendTelemetryEvent('cell.startExecute')
-    getEventReporter().reportExecution(cell)
+    await getEventReporter().reportExecution(cell)
```
jlewi, line 695: @sourishkrout Won't this block execution below? For example, suppose the Foyle service is down. We might reasonably expect the eventReporter to buffer the events and perform some retrying. But I don't want to block actual cell execution. Is there a better pattern to achieve this?

sourishkrout: Yes. I incorrectly assumed the noop event reporter would be unless the "AI experiment" is turned on.

I agree with you about this. I figured there are likely a few things that need to be addressed to move to a smooth "out-of-the-box" experience.

We could rewrite the Reporter to resemble the telemetry one (don't love the impl either actually), where it separates producing (enqueuing) and reporting the events. However, I'd suggest that for the time being, we use the noop logger unless AI is turned on. And track a separate effort to make this default robust.

I can rewrite this reasonably quickly at a later point.

jlewi: I decided to reply on the other PR
https://github.com/stateful/vscode-runme/pull/1589/files#r1739511628

PR #1625, opened 2024-09-06
Splits terminal buttons

pastuxso:
Splits terminal buttons to have separate business logic and conditions.

### DIFF
--- a/src/client/components/terminal/share.ts
+++ b/src/client/components/terminal/share.ts
@@ -8,13 +8,22 @@ import { SaveIcon } from '../icons/save'
 @customElement('share-cell')
 export class ShareCell extends LitElement {
   @property({ type: String })
-  shareText: string = 'Copy'
+  text: string = 'Copy'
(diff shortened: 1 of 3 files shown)


src/client/components/terminal/share.ts
```diff
@@ -8,13 +8,22 @@ import { SaveIcon } from '../icons/save'
 @customElement('share-cell')
```
sourishkrout, line 8: Is this still an appropriate name for the component?

sourishkrout (approved):
✅ LGTM. Just one nit, perhaps renaming the component?

src/client/components/terminal/share.ts
pastuxso: Sure, I think it should be more generic. Let me rename it.

sourishkrout: Something like `ActionButton` might be a good choice

PR #1628, opened 2024-09-10
Login prompt for Runme extension

pastuxso:
This PR enables the Stateful Authentication for Runme Extension, it will be prompted on-demand when a user hits the Save/Share button.


package.json
```diff
@@ -831,13 +842,13 @@
...
             "type": "string",
-            "default": "",
+            "default": "platform.stateful.com",
             "scope": "window",
             "markdownDescription": "Base domain to be use for Runme app"
           },
           "runme.app.panel.runme.cloud": {
             "type": "string",
-            "default": "runme.placeholder",
+            "default": "runme.cloud",
```
sourishkrout, line 851: We should give the panel its view and not reuse `runme.cloud`. Let's have `runme.optin` and have it alias `runme.placeholder` for now.

sourishkrout:
There's a way to not request a login as a badge as part of the Auth APIs and only do it if `ForceLogin` is enabled.

image

sourishkrout:
If I hit a `Save` button unauthed the first subsequent saving fails:

image

sourishkrout (changes requested):
See comments please. Might be worth to introduce the Opt-In (`runme.optin`) view first so we can hand it off to the team.

PR #1638, opened 2024-09-10
Await reporter after running the cell

sourishkrout:
This won't detach producer from consumer/reporter yet. However, it makes reporting more asynchronous, which is desirable.


sourishkrout:
Just a minor tweak, which should go a long way.

jlewi (approved):

Issue #1557, opened 2024-08-19
Refactor platform features to be opt-in for Runme but default for Stateful


sourishkrout:
The idea is simple. If logged in, irrespective of the extension, all features should be enabled (Save/Esclate/Open button, Auto-Save, etc). However, in Runme login should be strictly opt-in and not required.

It might be wise to refactor how some features are currently toggled. It'd be great to decouple visibility and required to be logged in so that it's possible to individually change the defaults for both Runme and Stateful (platform) extensions.

Happy to talk about this more, @pastuxso since the requirements are high-level.

PR #1642, opened 2024-09-10
Initial Feature flag config

pastuxso:
This approach utilizes the package.json file as a centralized location for adding or migrating features, including the respective conditions for rendering and simplifying overriding. 

This proposal aims to improve the separation of data/state in the Extension host and the Notebook Output UI. The Components can access the Features State through the Notebooks Communication API ([postKernelMessage and onDidReceiveKernelMessage](https://code.visualstudio.com/api/extension-guides/notebook#interactive-notebooks-communicating-with-the-controller))

Additionally, this PR enables the Stateful Authentication for Runme Extension, it will be prompted on-demand when a user hits the Save/Share button.

Closes: #1557


package.json
```diff
@@ -56,6 +56,56 @@
     "onStartupFinished"
   ],
   "main": "./out/extension.js",
+  "runme": {
+    "features": [
+      {
+        "name": "ForceLogin",
```
sourishkrout, line 62: Let's turn this array with a `name` per item into a object that uses the name as key. Something like this:

```
  "runme": {
    "features": {
      "forceLogin": {
        "enable": true,
        "conditions": {
          "os": "All",
          "vsCodeVersion": ">=1.58.0",
          "allowedExtensions": [
            "stateful.platform"
          ]
        }
      },
      "escalate": {
        "enable": true,
        "conditions": {
          "os": "All",
          "vsCodeVersion": ">=1.58.0",
          "statefulAuthRequired": true,
          "allowedExtensions": [
            "stateful.platform"
          ]
        }
      },
      "share": {
        "enable": true,
        "conditions": {
          "os": "All",
          "vsCodeVersion": ">=1.58.0",
          "allowedExtensions": [
            "stateful.platform",
            "stateful.runme"
          ]
        }
      },
      "gist": {
        "enabled": true,
        "conditions": {
          "os": "All",
          "vsCodeVersion": ">=1.58.0",
          "allowedExtensions": [
            "stateful.runme"
          ]
        }
      }
    }
  },
```

package.json
```diff
@@ -56,6 +56,56 @@
...
+        "enable": true,
+        "conditions": {
+          "os": "All",
+          "vsCodeVersion": ">=1.58.0",
+          "allowedExtensions": [
```
sourishkrout, line 67: We could also turn `allowedExtensions` into `enabledForExtensions` and then make the keys the extension handle and its value a true or false boolean. Something like this:

```
      "gist": {
        "enabled": true,
        "conditions": {
          "os": "All",
          "vsCodeVersion": ">=1.58.0",
          "enabledForExtensions": {
            "stateful.runme": true,
            "stateful.platform": false
          }
        }
      }
```

Again, the goal being that overwrites are an exception and not part of packaging the extension.

package.json
```diff
@@ -862,6 +901,30 @@
             "scope": "window",
             "default": "https://docs.runme.dev",
             "markdownDescription": "Documentation Base URL"
+          },
+          "runme.features.escalate": {
```
sourishkrout, line 905: I'd remove all these and generically handle settings lookup inside the code `runme.features.<feature-key` without publishing details here in `package.json` as a soft-overwrite for devs, not users.

.github/scripts/overwrites/stateful.sh
```diff
@@ -7,11 +7,13 @@ npm pkg set description="DevOps Notebooks built on Runme, connected for collabor
 npm pkg set homepage="https://stateful.com"
 npm pkg set contributes.configuration[0].properties[runme.app.baseDomain].default="platform.stateful.com"
 npm pkg set contributes.configuration[0].properties[runme.app.platformAuth].default=true --json
-npm pkg set contributes.configuration[0].properties[runme.app.enableShare].default=true --json
-npm pkg set contributes.configuration[0].properties[runme.experiments.escalationButton].default=true --json
 npm pkg set contributes.configuration[0].properties[runme.server.lifecycleIdentity].default=1 --json
 npm pkg set contributes.configuration[0].properties[runme.app.notebookAutoSave].default="yes"
 npm pkg set contributes.configuration[0].properties[runme.app.panel.runme.cloud].default="\"\"" --json
+npm pkg set contributes.configuration[0].properties[runme.features.forceLogin].default=true --json
```
sourishkrout, line 13: as per suggestion below - with feature flagging overwrites should be an exception and not part of the packaging process

src/extension/extension.ts
```diff
@@ -341,26 +342,6 @@ export class RunmeExtension {
       ),
     )
 
-    if (isPlatformAuthEnabled()) {
-      context.subscriptions.push(new StatefulAuthProvider(context, uriHandler))
-      const session = await getPlatformAuthSession(true)
-      if (session) {
-        const openDashboardStr = 'Open Dashboard'
-        const answer = await window.showInformationMessage(
```
sourishkrout, line 349: This `await` actually turns out to be permanently blocking until users respond to a non-modal notification. That's bad!

Can we fix this code to use the promise's then (as per this PR) to unblock completing initialization in a high priority separate PR @pastuxso? A user's action can't ever block the init method!

sourishkrout (changes requested):
See comments please.

sourishkrout (changes requested):
Also, now the badge no longer shows for `Stateful` which wasn't the requirement. It's not supposed to show for `Runme`.

package.json
```diff
@@ -56,6 +56,52 @@
...
   "main": "./out/extension.js",
+  "runme": {
+    "features": {
+      "ForceLogin": {
+        "enabled": true,
+        "conditions": {
+          "os": "All",
+          "vsCodeVersion": ">=1.58.0",
+          "enabledForExtensions": [
+            "stateful.platform"
```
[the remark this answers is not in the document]
pastuxso: @sourishkrout I see your point, I agree. I've been thinking more in the simple way to match the extension IDs, but it will be more difficult for overrides purposes. Likewise, I'll follow your suggestion.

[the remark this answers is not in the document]
sourishkrout: I think there's some beauty in explicitly expressing true and false versus using the absence. It's a matter of preference, I suppose.

pastuxso:
> Also, now the badge no longer shows for `Stateful` which wasn't the requirement. It's not supposed to show for `Runme`.

@sourishkrout, feel free to check it out if you want, it has been fixed.

I'll request you a new review round when I finish the changes with the `enabledForExtensions` to the key/value approach.

sourishkrout (approved):
✅ LGTM

PR #1644, opened 2024-09-11
Fix toggling between markdown/text with action buttons

sourishkrout:
It is likely broken due to the deprecation/changing of APIs. Require a minimum of VS Code v1.78.x now.

image


sourishkrout:
Please review, @pastuxso. It is currently broken in `stable`.

pastuxso (approved):
LGTM 🚀 ✅

PR #1656, opened 2024-09-12
Force Session Outputs in Platform Auth

peraltafederico:
Fix the issue where the session was not being sent to the Platform. This change forces session outputs to be cached when Platform Authentication is enabled, even if autosave is turned off.

It was necessary to change the encapsulation of the Feature Flagging functions to make it more readable and replicable in other parts of the code.

A new Flag called SignedIn was added to improve the conditions checking.

Using the session outputs cache to generate Gists was explored but was discarded since this would imply even more changes. It is proposed to do this functionality in another PR to continue as it works at this time in the stable version.

### DIFF
--- a/src/extension/cell.ts
+++ b/src/extension/cell.ts
@@ -50,7 +50,6 @@ import {
 } from './terminal/terminalState'
 import ContextState from './contextState'
 import { IRunnerEnvironment } from './runner/environment'
-import { GrpcSerializer } from './serializer'
 
 const NOTEBOOK_SELECTION_COMMAND = '_notebook.selectKernel'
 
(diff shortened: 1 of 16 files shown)


src/extension/extension.ts
```diff
@@ -203,8 +197,6 @@ export class RunmeExtension {
 
     await resetNotebookSettings()
 
-    const transientOutputs = !getSessionOutputs()
```
pastuxso, line 206: It seems that for the current usage of cache + in memory session outputs, it should be always `transientOutputs: false.`

src/extension/cell.ts
```diff
@@ -524,7 +526,9 @@ export class NotebookCellOutputManager {
    *
    */
   async refreshTerminal(terminalState: ITerminalState | undefined): Promise<void> {
-    if (!ContextState.getKey(NOTEBOOK_AUTOSAVE_ON)) {
+    const isSignedIn = features.isOnInContextState(FeatureName.SignedIn)
+
+    if (!isSignedIn) {
```
pastuxso, line 531: It should be validated only for Stateful, because when you turn on the reporter experiment it's not able to capture the exitCode.

sourishkrout (approved):
LGTM ✅. Based on testing both extension, I believe we have parity with pre-feature-flagging now.

sourishkrout:
Liking how the features interface is a lot more narrow now 👍.

PR #1661, opened 2024-09-14
Remove option to enable aiLogs in the server

jlewi:
* This effectively reverts https://github.com/stateful/vscode-runme/pull/1380
* Per https://github.com/jlewi/foyle/issues/211 we are no longer relying on RunMe logs
* Rather RunMe now uses Foyle APIs to send events to Foyle as necessary.


sourishkrout (approved):
✅  LGTM

PR #1662, opened 2024-09-16
Fix code-server login

peraltafederico:
This resolves the issue in code-server where a `Couldn't recognize command` error occurs during authentication against the Platform, because it expects the `windowId` URI parameter, which is missing.

The fix has been successfully tested in both code-server and standard VSCode environments.

### DIFF
--- a/src/extension/handler/uri.ts
+++ b/src/extension/handler/uri.ts
@@ -57,7 +57,6 @@ export class RunmeUriHandler implements UriHandler, Disposable {
   async handleUri(uri: Uri) {
     log.info(`triggered RunmeUriHandler with ${uri}`)
     const params = new URLSearchParams(uri.query)
-    const windowId = params.get('windowId')
     const state = params.get('state')
     const code = params.get('code')
     const command = params.get('command')
@@ -66,7 +65,7 @@ export class RunmeUriHandler implements UriHandler, Disposable {
       window.showErrorMessage('No query parameter "command" provided')
       return
     }
-    if (command === 'auth' && windowId && state && code) {
+    if (command === 'auth' && state && code) {
       TelemetryReporter.sendTelemetryEvent('extension.uriHandler', {
         command,
         type: AuthenticationProviders.Stateful,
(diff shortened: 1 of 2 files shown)


sourishkrout:
I'm wondering if `windowId` is something one can only expect from desktop (electron app) distributions of VS Code?

peraltafederico:
> I'm wondering if `windowId` is something one can only expect from desktop (electron app) distributions of VS Code?

@sourishkrout It is definitely related to letting VSCode handle links in different or existing windows, as mentioned here: https://code.visualstudio.com/updates/v1_67#_handle-uri-in-new-window. However, I didn’t find any explicit information stating that this behavior is limited to desktop distributions only.

That said, `windowId` is not related to the authentication process, so I think it's safe to remove it.

sourishkrout:
> That said, `windowId` is not related to the authentication process, so I think it's safe to remove it.

I'd like to hear @pastuxso's opinion. Also, why not check if it's defined and set it accordingly but not require it in the condition? Again, I haven't checked, but it seems important when multiple Windows (desktop apps) are running.

peraltafederico:
> > That said, `windowId` is not related to the authentication process, so I think it's safe to remove it.
> 
> I'd like to hear @pastuxso's opinion. Also, why not check if it's defined and set it accordingly but not require it in the condition? Again, I haven't checked, but it seems important when multiple Windows (desktop apps) are running.

Totally! :100:

Regarding the condition and how the parameter is handled, it's only used as a validation to define if the URI is valid or not. The whole URI is then forwarded to the next handler, see https://github.com/stateful/vscode-runme/pull/1662/files#diff-35ad708e91026099e03a5b146676d059a44e4aab20fd04d76412c9c485ec73c4R73.

sourishkrout:
That sounds good, @peraltafederico. Since @pastuxso wrote this code initially, I assigned him as a reviewer. I'm using a build from this branch to unblock my development work. It'd be great to get his fixed soon but it's no emergency.

pastuxso:
@sourishkrout The validation of the **windowId** parameter can be safely removed. Initially, it was used to differentiate between commands, but with the introduction of the argument **command=auth**, it has become redundant.

pastuxso (approved):
LGTM 🚀

PR #1670, opened 2024-09-18
Removes vsCodeVersion to skip checkVersion

pastuxso:
At the moment, the feature flag functionality is checking the VSCode version to enable/disable a feature. For now this PR removes this condition to avoid conflicts with other flavors like VSCode Insiders, VSCodium, Cursor, etc.


sourishkrout (approved):
✅ LGTM

PR #1672, opened 2024-09-19
Loads StatefulAuth provider only when it's necessary

pastuxso:
The Runme extension loads the StatefulAuth provider, which could lead to a poor user experience. Even when the silent flag is passed, there is a listener that may activate the Badge in the User Menu, prompting a login suggestion.

In future PRs, when support for saving cells on the Platform is introduced, this flag can be enabled for Runme.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,10 +58,18 @@
   "main": "./out/extension.js",
   "runme": {
     "features": {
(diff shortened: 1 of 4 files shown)


sourishkrout (approved):
✅ LGTM

PR #1679, opened 2024-09-20
Enables Cells Saving in Runme

pastuxso:
Enables support to Save Cells in the Stateful Platform.

Updates the validation that skips writing session file, now it considers if auto-save is on/off to resolve it.

In Runme the Lifecycle ID is configured by default at Cell level, so the experiment **reporter** `(runme.experiments.reporter)` fails due to it needs the Frontmatter, which is not available. To solve this issue, this PR uses the Ephemeral ID as fallback.

#### Runme extension known bugs:

- Dagger terminal outputs are not working with the Save button. I disable it for this kind of outputs.

### DIFF
--- a/src/extension/messages/platformRequest/saveCellExecution.ts
+++ b/src/extension/messages/platformRequest/saveCellExecution.ts
@@ -190,8 +190,12 @@ export default async function saveCellExecution(
       let fmParsed = editor.notebook.metadata['runme.dev/frontmatterParsed'] as Frontmatter
 
       if (!fmParsed) {
-        const yamlDocs = YAML.parseAllDocuments(editor.notebook.metadata['runme.dev/frontmatter'])
-        fmParsed = yamlDocs[0].toJS?.()
+        try {
+          const yamlDocs = YAML.parseAllDocuments(editor.notebook.metadata['runme.dev/frontmatter'])
+          fmParsed = yamlDocs[0].toJS?.() || {}
+        } catch (e) {
+          log.warn('failed to parse frontmatter')
+        }
       }
 
       let notebookInput: CreateNotebookInput | undefined
--- a/src/extension/messages/platformRequest/saveCellExecution.ts
+++ b/src/extension/messages/platformRequest/saveCellExecution.ts
@@ -52,7 +52,6 @@ export default async function saveCellExecution(
   }
 
   const cacheId = GrpcSerializer.getDocumentCacheId(metadata) as string
-  console.log('saveCellExecution:CacheId: ', cacheId)
   const plainSessionOutput = await kernel.getPlainCache(cacheId)
   const maskedSessionOutput = await kernel.getMaskedCache(cacheId)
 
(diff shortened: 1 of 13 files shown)


sourishkrout:
> Runme extension known bugs
> - When a file is crated and a Cell is inserted and run, it is not sending the file contents to the Platform. It requires re-opening the file to work as it's expected. I'll face it in a separate branch.

While it's probably safe to merge this branch without enabling the `Save` button, I think we need a solution for the abovementioned issue. For the dagger integration we could just disable the Save button or whatever is the most straightforward.

sourishkrout (changes requested):
See comments.

src/extension/serializer.ts
```diff
@@ -834,12 +823,21 @@ export class GrpcSerializer extends SerializerBase {
...
         version?: string
       }
+    } = { runme: {} }
+
+    if (rawFrontmatter) {
+      try {
+        const yamlDocs = YAML.parseAllDocuments(metadata['runme.dev/frontmatter'])
+        data = (yamlDocs[0].toJS?.() || {}) as typeof data
+      } catch (e) {
+        log.warn('failed to parse frontmatter')
```
sourishkrout, line 839: What about the error itself? Why not included it in the logging?

src/extension/messages/platformRequest/saveCellExecution.ts
```diff
@@ -91,10 +91,21 @@ export default async function saveCellExecution(
         marshalFrontmatter: true,
       })
 
+      // Fallback to Ephemeral ID if there is no frontmatter
+      if (!notebook?.frontmatter?.runme?.id) {
```
sourishkrout, line 95: Won't using an ephemeral ID undermine the ability to track a document reliably?

I think the solution is to tie the lifecycle ID setting to the relevant features. In other words, document and cell IDs should be written if Stateful features are in play.

sourishkrout (changes requested):
See comments.

src/extension/serializer.ts
pastuxso: Good point. I missed including it.

src/extension/messages/platformRequest/saveCellExecution.ts
pastuxso: For Stateful extension, It's not a problem right now because the default settings. 

What you say is to update settings like **Auto-Save** to turn on Lifecycle ID if the Save button is clicked?

sourishkrout: > What you say is to update settings like Auto-Save to turn on Lifecycle ID if the Save button is clicked?

Yes. Just like all the other former settings, now features, even lifecycle identity is no longer a build-time setting when we blur the lines between the two extension bundles.

admc: What will happen if they manually turn of lifecycle but are authed with Stateful?

pastuxso: @admc Right now with this PR, it will take the Ephemeral one until a session ends or the file reopens.

sourishkrout (approved):
Approved ✅ pending the merge of https://github.com/stateful/vscode-runme/pull/1694.

PR #1683, opened 2024-09-24
Finalize Runme terminal

sourishkrout:
Adds Runme terminal in two places:

1. Cell's `Fork` button.
2. Pick `Runme Terminal` inside the `+` menu in the `Terminal` panel.

It supports rolling up ENV vars in the current/forked session. When exited by typing `save`, a cell with all the terminal sessions will be added for 1.), and a new notebook document will be started for 2.

It only works in tandem with https://github.com/stateful/runme/pull/673.


pastuxso (approved):
LGTM 🚀

Issue #1580, opened 2024-08-21
Node 20 Support


pastuxso:
Recently, I noticed how our dependency modules are moving to Node 20. For example: https://github.com/microsoft/vscode-vsce/pull/1035 https://github.com/stateful/vscode-runme/pull/1567

Right now, the unique blocker is the usage of the **--experimental-specifier-resolution=node**, which starting from Node v19 has been deprecated. To replace it requires the usage of loaders like https://github.com/barhun/extensionless, which enables support to import modules without the file extension. In our case, modules from [Buf](https://buf.build/) are using relative imports without file extensions.

sourishkrout:
More importantly we should probably track with VS Code's Electron/Node version:

```
Version: 1.92.2 (Universal)
Commit: fee1edb8d6d72a0ddff41e5f71a671c23ed924b9
Date: 2024-08-14T17:29:30.058Z
Electron: 30.1.2
ElectronBuildId: 9870757
Chromium: 124.0.6367.243
Node.js: 20.14.0
V8: 12.4.254.20-electron.0
OS: Darwin arm64 23.6.0
```

PR #1696, opened 2024-09-27
Bump to Node v20.17.0

pastuxso:
Updates to Node v20.17.0.

Replaces `--experimental-specifier-resolution=node` by an integrated loader `--import=./specifier-register.mjs` based on the [extensionless](https://github.com/barhun/extensionless) loader.

I couldn't use the **Extensionless** module due to issues during the e2e test run, It seems to be related with the usage of the **NODE_OPTIONS** env variable, once it is established all scripts inherits its value, and sometimes it is not required, i.e. a **npm install**. This is the reason why I created a symlink in the directory **tests/e2e**.

Closes: https://github.com/stateful/vscode-runme/issues/1580


sourishkrout (approved):
✅ LGTM

PR #1700, opened 2024-10-02
Add support to perform copy to clipboard on Linux

pastuxso:
On Linux, the Notebook Terminal, it doesn't have the native capability to allow copy the selected text. Only works using the third mouse button to paste, I think this is a behavior of the X Server-Client.

This PR includes changes to allow copy to the clipboard using the most used keyboard shortcut, which is `SHIFT+CTRL+C`. It is designed to avoid conflicts with the `CTRL+C` (Terminate process).

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,16 @@
   "main": "./out/extension.js",
   "runme": {
     "features": {
+      "CopySelectionToClipboard": {
+        "enabled": true,
+        "conditions": {
+          "os": "linux",
+          "enabledForExtensions": {
+            "stateful.platform": true,
+            "stateful.runme": true
+          }
+        }
+      },
       "RequireStatefulAuth": {
         "enabled": true,
         "conditions": {
--- a/src/client/components/terminal/index.ts
+++ b/src/client/components/terminal/index.ts
@@ -442,6 +442,17 @@ export class TerminalView extends LitElement {
     this.terminal.loadAddon(this.serializer)
     this.terminal.loadAddon(new Unicode11Addon())
     this.terminal.loadAddon(new WebLinksAddon(this.#onWebLinkClick.bind(this)))
+
+    this.terminal.attachCustomKeyEventHandler((event) => {
+      if (event.shiftKey && event.ctrlKey && event.key === 'C') {
+        const selection = this?.terminal?.getSelection()
+        if (selection) {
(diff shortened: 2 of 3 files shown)


sourishkrout (approved):
✅ LGTM

forseti:
Hi Team,
I've just tested this new change in `3.8.6`

The copy feature still does not copy the selection.

### Edit > Copy 
This copies the content of script instead of the selection in the cell output

sourishkrout:
Please look into this @pastuxso.

forseti:
Hi @pastuxso @sourishkrout 

I just retested `Shift+Ctrl+C` again and found that it actually copied the selection correctly. 
Although on the downside, it also opens the browser's `Developer Tools` or VSCode's Console,

Thanks.

pastuxso:
Hi @forseti, Just to confirm, are you using interactive terminal?

image

forseti:
Hi @pastuxso 
Yes, the Copy actually works now after I disabled the interactive terminal.
Thanks!

pastuxso:
@forseti No problem. But, the fix is expected to work for interactive terminals. Let me figure out, thank you for your feedback.

pastuxso:
@forseti Confirmed. Unfortunately, the Contextual COPY is not taking the output content, and For web VS Code Flavors, the shortcut is in conflict with the Developer Tools One. 

Our approach works only for Desktop Installations. I'll work on my current priorities, after that I'll fix it. I'll be back with you as soon as possible.

Meanwhile, no-interactive terminals can unblock you, or using a Desktop version using the shortcut SHIFT+CRL+C to copy the selected text.

forseti:
Hi @pastuxso

> Confirmed. Unfortunately, the Contextual COPY is not taking the output content, and For web VS Code Flavors, the shortcut is in conflict with the Developer Tools One.
> Our approach works only for Desktop Installations. I'll work on my current priorities, after that I'll fix it. I'll be back with you as soon as possible.

Okay. No problem. :+1:

> Meanwhile, no-interactive terminals can unblock you, or using a Desktop version using the shortcut SHIFT+CRTL+C to copy the selected text.

Thanks again for the help! :pray:

I wasn't aware of the non-interactive mode before. Now I can always have my cell configured with `"interactive":"false"` if I need to copy their outputs.

PR #1709, opened 2024-10-02
Include selected index in GenerateCellsRequest

jlewi:
* Related to jlewi/foyle#111
* The frontend needs to send the index of the selected cell to the AI in order for the AI to properly generate the completion.

* Without the cell index the AI would just keep generating starting at the end of the document which produces bad results when an earlier cell is selected.

* This requires updating the Foyle client library to pull in a version of the protos which includes selected_index in the request


sourishkrout (approved):
✅ LGTM

PR #1712, opened 2024-10-07
Minor Fixes and Enhancements

pastuxso:
- [x] Fix: Corrected Gist filename handling when it contains hyphens.
- [x] Fix: Resolved crash on Linux when executing `runme run npm-watch`.
- [x] Fix: Glob pattern to allow **.mdx**, and **.mdr** files
- [x] Enhancement: Initial terminology update by renaming **categories** to **tags** in the UI. For a full change, it will require changes in the **Graphql** Schema too. Because, the Category annotation is fully integrated with the Cloud.
- [x] Enhancement: Opens Runme Cloud Panel instead of Force Login for Runme on Saving cells.

### DIFF
--- a/src/extension/provider/launcher.ts
+++ b/src/extension/provider/launcher.ts
@@ -37,7 +37,7 @@ interface TreeFile {
  */
 let i = 0
 
-const GLOB_PATTERN = '**/*.{md,mdr,mdx}'
+export const GLOB_PATTERN = '**/*.{md,mdr,mdx}'
 
 export class RunmeFile extends TreeItem {
   constructor(
--- a/src/extension/provider/launcher.ts
+++ b/src/extension/provider/launcher.ts
@@ -37,6 +37,8 @@ interface TreeFile {
  */
 let i = 0
 
+const GLOB_PATTERN = '**/*.{md,mdr,mdx}'
+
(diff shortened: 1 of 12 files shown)


src/extension/provider/launcher.ts
```diff
@@ -37,6 +37,8 @@ interface TreeFile {
  */
 let i = 0
 
+export const GLOB_PATTERN = '**/*.{md,mdr,mdx}'
```
sourishkrout, line 40: @pastuxso what about `.runme`, `.run`, `.mdi`?

https://github.com/stateful/vscode-runme/blob/main/package.json#L595-L613

sourishkrout:
> * For a full change, it will require changes in the **Graphql** Schema too. Because, the Category annotation is fully integrated with the Cloud.

Limiting the change to the UI/UX labels is fine for now 👍.

sourishkrout (changes requested):
> [x] Enhancement: Opens Runme Cloud Panel instead of Force Login for Runme on Saving cells.

This isn't working for me. Does it work reliably for you? I recall there were limitation to open/jump to panels. I might be mistaken but something about that you can only open them if they were opened at least once before during the session.

sourishkrout (approved):
✅ LGTM

PR #1726, opened 2024-10-09
Dependabot 20241009

pastuxso:
Dependabot updates for 20241009. 
 
The dependency **@types/node-fetch** was added manually, due to it has been taken from **Vercel** before.


sourishkrout:
> The dependency **@types/node-fetch** was added manually, due to it has been taken from **Vercel** before.

Do we still have a `vercel` dependency? We can probably remove it since we've deprecated the Vercel integration.

pastuxso:
@sourishkrout Yes, we do. I believe it would be better to remove it in a separate pull request.

sourishkrout (approved):
✅ LGTM

PR #1733, opened 2024-10-15
Explicit errors when saving cells 

peraltafederico:
Instead of casting, explicitly throw an error if the notebook cache or the cell the user is trying to save is not found.


sourishkrout (changes requested):
@peraltafederico no actual change left, only tests.

peraltafederico:
> @peraltafederico no actual change left, only tests.

@sourishkrout  Sorry, this is embarrasing. I must have force-pushed without a stale version of `saveCellExecution.ts` or something. I just pushed the correct commit :+1: .

sourishkrout (approved):
✅ LGTM

PR #1745, opened 2024-10-18
Report cell execution status to Foyle.

jlewi:
* This updates the extension to report the status of the cell execution to Foyle when reporting execution events.

* This is highly valuable signal for the AI. This will help us from learning from broken/invalid commands.

* Fix jlewi/foyle#310


jlewi:
I think there's more work to be done here. It looks like this misses the outputs of the execution in some cases.

jlewi:
@sourishkrout this is ready for review.

sourishkrout (commented):
✅ LGTM

PR #1755, opened 2024-10-23
Version 0

hotpocket:
Exploring modifying runme.


package.json
```diff
@@ -677,9 +677,15 @@
           "runme.experiments.grpcSerializer": {
             "type": "boolean",
             "scope": "window",
-            "default": true,
+            "default": false,
             "markdownDescription": "If set to `true`, the notebook serialization will use the GRPC instead of the WebAssembly interface."
           },
+          "runme.experiments.connectSerializer": {
```
jlewi, line 683: Can we make this a string serializerAddress?
https://docs.google.com/document/d/1blm5kD0rb0X6hyb3w40gwYrhKzeCVbxGwLJUb0vkCCo/edit#bookmark=id.tdj09kndh2f6

src/extension/extension.ts
```diff
@@ -127,11 +128,13 @@ export class RunmeExtension {
     )
 
     const reporter = new GrpcReporter(context, server)
-    const serializer = grpcSerializer
-      ? new GrpcSerializer(context, server, kernel)
-      : new WasmSerializer(context, kernel)
+    const serializer = useConnectSerializer
```
jlewi, line 131: Using the "?" : with nested  "?": isn't very readable.
I think an if/else would be more readable.

Per the comment above the argument for the connectSerializer option should be a string containing the address. You should use non-empty string to indicate that connectSerializer should be used. Otherwise you should fall back to the existing logic.

PR #1757, opened 2024-10-24
Add a Foyle status bar in the lower right corner

jlewi:
* Use the status bar to display the session id
* The session id is used to track all the AI generations associated with a particular context

* Surfacing the session id makes it easy to fetch the relevant Foyle logs to understand why the AI made a particular suggestion.

Fix: jlewi/foyle#322


sourishkrout (approved):
✅ LGTM

PR #1759, opened 2024-10-25
Adds authentication support for the extension in playground environments

pastuxso:
Adds authentication support for the extension in playground environments, integrating cloud token validation and enabling token retrieval from callback URLs dispatched by the cloud. Key changes include:

- **Auth Token Handling:** Simplified by removing unused refresh token logic, as the access-token has an extended lifespan.
- **Load from Secrets:** An instance method now checks tokens against playground secrets, using the cloud endpoint for validation.
- **Callback Integration:** Supports token recovery from cloud callback URLs, enhancing seamless access when the user is already Authenticated in the Cloud.


src/extension/provider/statefulAuth.ts
```diff
@@ -248,6 +237,71 @@ export class StatefulAuthProvider implements AuthenticationProvider, Disposable
     this.#disposables.forEach((d) => d.dispose())
   }
 
+  public async loadFromSecrets() {
```
sourishkrout, line 240: It's a `token`. More specifically, "secrets" is too vague.

Also, perhaps it should be `loginFromToken` or `bootstrapFromToken` because it does not "just" do loading.

src/extension/provider/statefulAuth.ts
```diff
@@ -248,6 +237,71 @@ export class StatefulAuthProvider implements AuthenticationProvider, Disposable
...
+    const secretsFile = '/etc/secrets/playground-dev'
```
sourishkrout, line 241: It should be overwritable by a setting. `/etc...` does not sound on non-Linux file systems. In other instances, we use the extension directory available in `ExtensionContext` to ensure it is always readable, irrespective of where the extension runs. The dir name should reflect what its purpose is and not its application. The word playground is irrelevant here. Something like `./secerts/authToken` is a good choice.

We should have a separate setting (default true) to toggle whether or not the token should be deleted once used.

src/extension/provider/statefulAuth.ts
```diff
@@ -248,6 +237,71 @@ export class StatefulAuthProvider implements AuthenticationProvider, Disposable
...
+      return
+    }
+
+    const rawToken = await workspace.fs.readFile(secretsUri)
+    if (!rawToken?.length) {
+      logger.error('Failed to read secrets file')
+      return
+    }
+
+    const token = new TextDecoder().decode(rawToken).trim()
```
sourishkrout, line 259: 259-270 should go into a protected/private method `InsecureDecode` that's merely being called here. The code does not indicate by itself that the JWT is *not* being verified until the API is called further down.

src/extension/provider/statefulAuth.ts
```diff
@@ -248,6 +237,71 @@ export class StatefulAuthProvider implements AuthenticationProvider, Disposable
...
+      return
+    }
+
+    const { exp, scope } = jwtDecoded
+    if (!exp || !scope) {
+      logger.error('Invalid JWT token')
+      return
+    }
+
+    try {
```
sourishkrout, line 272: Let's break out the code into a private/protected method that makes the authentication decision and name the method accordingly.

Having less detail right here and calling into subroutines with distinct names will clarify that.

sourishkrout (changes requested):
Works as advertised ✅.

However, let's refactor the code according to my review and, if possible, let's add a simple unit test which should be easier once decoding and auth are broken out.

src/extension/provider/statefulAuth.ts
```diff
@@ -248,6 +237,90 @@ export class StatefulAuthProvider implements AuthenticationProvider, Disposable
...
+      logger.info('No auth token file found, halting bootstrap from token.')
+      return
+    }
+
+    try {
+      const { token, payload } = await this.insecureDecode(authTokenUri)
+      const session = await this.buildSession(token, payload)
+      await this.persistSessions([session], { added: [session], removed: [], changed: [] })
+      await this.deleteAuthTokenFile(authTokenUri)
+      return true
```
sourishkrout, line 253: Return true here but `undefined` everywhere else. I suggest adding a `: boolean` to the return type of the function and making sure it's consistent.

Semantically, undefined and false are the same but it's sloppy typing.

src/extension/provider/statefulAuth.ts
```diff
@@ -248,6 +237,90 @@ export class StatefulAuthProvider implements AuthenticationProvider, Disposable
     this.#disposables.forEach((d) => d.dispose())
   }
 
+  public async bootstrapFromToken() {
+    const authTokenUri = Uri.joinPath(this.context.extensionUri, 'secrets', 'authToken')
```
sourishkrout, line 241: As said in the first review, the location of the token should be configurable and default to the location hardcoded here.

src/extension/extension.ts
```diff
@@ -451,8 +451,11 @@ export class RunmeExtension {
     }
 
     if (kernel.isFeatureOn(FeatureName.RequireStatefulAuth)) {
-      const forceLogin = kernel.isFeatureOn(FeatureName.ForceLogin)
-      context.subscriptions.push(new StatefulAuthProvider(context, uriHandler))
+      const statefulAuthProvider = new StatefulAuthProvider(context, uriHandler)
+      context.subscriptions.push(statefulAuthProvider)
+
+      const secretsAuthed = await statefulAuthProvider.bootstrapFromToken()
```
sourishkrout, line 457: Should this run if one's already signed in? More so, should a token be processed if there's a valid session?

The problem is that if a token was previously used to successfully bootstrap but skipping token deletion is now invalid?

.gitignore
```diff
@@ -18,3 +18,4 @@ examples/.vscode
 assets
 .venv
 **/*-01*.md
+/secrets
```
sourishkrout, line 21: This also needs to go into `.vscodeignore` to exclude it from VSIX packaging

sourishkrout (changes requested):
Almost there. Please see comments.

src/extension/extension.ts
pastuxso: Solved. Now, I'm checking the sessions first.

sourishkrout (changes requested):
The path is not working because it's not supposed to concatenate to the extension base. Also, there's no default, which wasn't what I asked for.

sourishkrout (approved):
✅ LGTM

PR #1762, opened 2024-10-28
Dependabot config to pin `@types/node` to `v20`

pastuxso:
Pins the module `@types/node` to the v20.x in order to be consistent with the Node version.


sourishkrout (approved):
👍 LGTM

PR #1763, opened 2024-10-29
Rewire stdio for non-interactive

sourishkrout:
As @jlewi summarized in https://github.com/stateful/runme/issues/684:

> * Interactive terminals multiplex stdout and stderr and show the multiplexed stream in the Cell output window
> * non-interactive terminals don't multiplex stdout/stderr and only stdout is sent to the cell output window
> * non-interactive terminals don't multiple stdout & stderr because this could interfere with 3rd party renders
>   
>   * For example imagine a CLI is emitting a JSON table to stdout and also sending errors to stderr
>   * If you multiple stdout & stderr you no longer have a stream of valid JSON
>   * So a 3rd party renderer for JSON would break if you multiplexed stdout and stderr
> * So the proposed fix for non-interactive terminals is to send stderr to the vscode integrated terminal in the bottom panel

This PR aligns the user experience with the above. Todos:

- [x] Add/fix tests
- [x] Test dagger integration continues to work
- [x] Test cloud resource integration continues to work
- [x] Finalize stderr "color" if any, it's a less alarming terminal-yellow right now

Btw, this PR does not solve how to include `stderr` into the serialized notebook, as per:

> and for `non-interactive` we could probably do:
> 
> * `stateful.runme/output-items`
> * `application/vnd.code.notebook.stdout`
> * perhaps a third with `application/vnd.code.notebook.stderr`

We'll handle this in a separate PR (cc @jlewi).

### DIFF
--- a/src/client/components/configuration/annotations.ts
+++ b/src/client/components/configuration/annotations.ts
@@ -57,7 +57,10 @@ export class Annotations extends LitElement {
       description: 'Run cell inside terminal to allow for interactive input.',
     },
     closeTerminalOnSuccess: {
-      description: 'Hide terminal after cell successful execution.',
+      description: 'Hide terminal panel after cell successful execution.',
+    },
+    openTerminalOnError: {
+      description: 'open terminal panel after cell execution error.',
     },
     promptEnv: {
       description: 'Prompt user input for exported environment variables.',
--- a/src/extension/cell.ts
+++ b/src/extension/cell.ts
@@ -224,7 +224,6 @@ export class NotebookCellOutputManager {
         return new NotebookCellOutput([NotebookCellOutputItem.json(json, OutputType.vercel)])
       }
 
-      case OutputType.outputItems:
       case OutputType.terminal: {
         const terminalState = this.terminalState
         if (!terminalState) {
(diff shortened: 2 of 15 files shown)


sourishkrout:
@jlewi, here's a demo. You should be able to run it locally. Merging it still requires more work, though.

https://github.com/user-attachments/assets/89ee635f-5860-408a-8ee5-bc88ea465a9e

jlewi:
The UX looks great. One question I have for non-interactive cells, if I wanted to programmatically access stderr in order to send it to Foyle, where would I tap into it.

Consider the following example
1. We use gcloud to fetch logs from Cloud Logging
2. We use a custom renderer to render logs in a rich way
3. gcloud fails with an error because the query is wrong

In this case what I'd like to happen is
1. We send a request to Foyle with the stderr containing the error message about the logging query
2. Foyle returns a corrected command which is inserted as a ghost cell.

sourishkrout:
I see, @jlewi. We could employ the same strategy for `non-interactive`, which is already being used for `interactive`, where we create a single output with multiple items (different media types). This is how we render Gists with serialized output whereas the notebook shows a terminal.

For `interactive`, we have the following:
- `stateful.runme/terminal`
- `application/vnd.code.notebook.stdout`

and for `non-interactive` we could probably do:
- `stateful.runme/output-items`
- `application/vnd.code.notebook.stdout`
- perhaps a third with `application/vnd.code.notebook.stderr`

Or combine stdout+stderr into a single item.

sourishkrout:
@pastuxso please review.

pastuxso (approved):
LGTM 🚀

PR #1784, opened 2024-11-07
Rendering New Notebooks Panel using VS Code Task Provider

pastuxso:
Refactoring of New Notebooks Panel in order to avoid duplication and to have a unique source of truth for tasks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -427,6 +427,11 @@
         "description": "Open Cloud Panel"
       },
       {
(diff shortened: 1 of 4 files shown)


sourishkrout (commented):
Is it possible to not uncollapse the whole tree when switching between named / unnamed @pastuxso?

https://github.com/user-attachments/assets/6bb592c8-a8f9-4409-9d62-68b937782568

PR #1785, opened 2024-11-07
Wait for authentication to complete before you can save a cell

pastuxso:
This update ensures that users must complete the authentication process before saving a cell. It improves reliability by handling authentication state transitions seamlessly during cell run, enhancing the user experience.

It now makes its best effort to wait for the user to complete the happy path, keeping the save button in a loading state. If authentication is completed within a minute, the cell is saved automatically without requiring the user to press save again. If authentication fails, the user is notified, and the save button returns to its initial state.

- [x] Event to wait until login finishes. 
- [x] Handle login timeout.
- [x] Edge case: Running cells again while it is waiting for login.

https://github.com/user-attachments/assets/fe879d71-28d3-4b29-83e0-2252ec1bd6f7

### DIFF
--- a/__mocks__/vscode.ts
+++ b/__mocks__/vscode.ts
@@ -327,4 +327,9 @@ export enum EndOfLine {
   CRLF = 2
 }
 
+export class CancellationTokenSource {
+  cancel = vi.fn()
+  dispose = vi.fn()
+}
+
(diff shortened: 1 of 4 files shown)


sourishkrout (approved):
✅ LGTM

PR #1786, opened 2024-11-08
Improves user experience inside Stateful Cloud Playgrounds

pastuxso:
Introduces changes to enhance the user experience when running hosted extension within Stateful Cloud Playgrounds.

* Configure trusted domains.
* Auto open a new terminal only the first time.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,16 +59,7 @@
   "main": "./out/extension.js",
   "runme": {
     "features": {
-      "AutoOpenTerminal": {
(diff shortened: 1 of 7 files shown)


sourishkrout (changes requested):
@pastuxso can we move these two UX features into one that is `isHostedPlayground` or something like that please?

sourishkrout (approved):
LGTM ✅

PR #1788, opened 2024-11-13
Track Runme runs (logged into Stateful-only)

peraltafederico:
- Track runme cell run events and save it to the Cloud database only when the user is logged in.
- Some type narrowing clean up to avoid casting

Depends on https://github.com/stateful/platform/pull/1213

### DIFF
--- a/src/extension/kernel.ts
+++ b/src/extension/kernel.ts
@@ -84,7 +84,6 @@ import {
   getEnvProps,
   warnBetaRequired,
   getPlatformAuthSession,
-  isTelemetryEnabled,
 } from './utils'
 import { getEventReporter } from './ai/events'
 import { getSystemShellPath, isShellLanguage } from './executors/utils'
@@ -910,9 +909,7 @@ export class Kernel implements Disposable {
     })
     const endTime = Date.now()
 
-    if (isTelemetryEnabled()) {
-      this.#handleTrackCellRun(cell, successfulCellExecution, startTime, endTime)
-    }
+    this.#handleTrackCellRun(cell, successfulCellExecution, startTime, endTime)
 
     runmeExec.end(successfulCellExecution, endTime)
   }
--- a/src/extension/kernel.ts
+++ b/src/extension/kernel.ts
@@ -39,6 +39,7 @@ import {
   type ExtensionName,
   type FeatureContext,
   FeatureName,
+  APIMethod,
 } from '../types'
 import {
   ClientMessages,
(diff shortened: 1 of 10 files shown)


src/extension/kernel.ts
```diff
@@ -832,7 +901,11 @@ export class Kernel implements Disposable {
       'cell.success': successfulCellExecution?.toString(),
       'cell.mimeType': annotations.mimeType,
     })
-    runmeExec.end(successfulCellExecution, Date.now())
+    const endTime = Date.now()
+
+    this.#handleTrackCellRun(cell, successfulCellExecution, startTime, endTime)
```
pastuxso, line 906: @peraltafederico I think the call should be conditioned to the telemetry flag `isTelemetryEnabled`. Take a look at https://github.com/stateful/vscode-runme/commit/4c11a9c2dbaf8a56e9dcac07febbc5d2f7eae744#diff-2b0a484742f2baf7a6b33fec5aa69dfbf17fbe089363cec3549d0b88f664b1b4R23

src/extension/kernel.ts
```diff
@@ -734,6 +738,70 @@ export class Kernel implements Disposable {
...
 
+  async #handleTrackCellRun(
+    cell: NotebookCell,
+    successfulCellExecution: boolean,
+    startTime: number,
+    endTime: number,
+  ): Promise<void> {
+    const session = await getPlatformAuthSession(false)
+
+    if (!this.serializer?.isLifecycleIdentity(RunmeIdentity.ALL || !session)) {
```
pastuxso, line 749: I think the `||` should after closing the call `isLifecycleIdentity`.

```suggestion
    if (!this.serializer?.isLifecycleIdentity(RunmeIdentity.ALL) || !session) {
```

pastuxso: Apart of that, I think is not necessary to add a new method to the serializer. It's enough calling the runtime setting value directly from `ContextState.getKey<ServerLifecycleIdentity>(NOTEBOOK_LIFECYCLE_ID)`.

pastuxso (changes requested):

src/extension/kernel.ts
peraltafederico: Ha! Totally. Last minute change to make readability better and I missed that nonsense parenthesis. Good catch!

peraltafederico: Fixed in 99e68515e54b2cb2745ea9658518dfc16b5cd32e

src/extension/kernel.ts
peraltafederico: Fixed in 99e68515e54b2cb2745ea9658518dfc16b5cd32e

sourishkrout:
I'll look at it over the weekend.

sourishkrout (approved):
I want to detach rendering/packaging and delivering the message for the former to be sync and the latter to be async. Merging for now.

I will refactor this to be sync+async next week.

PR #1799, opened 2024-11-18
Fully async to never block execution

sourishkrout:
1. Break out code that only runs for logged in users
2. Enqueue cell runs (sync), immediately return to unblock execution
3. Report cell runs only for logged in users (async)


pastuxso (approved):
LGTM 👍

PR #1804, opened 2024-11-19
Recommend extensions sensitive to current repo

sourishkrout:
Only ask if a user has used Runme frequently enough (as per survey API).


pastuxso (approved):
LGTM 👍

PR #1805, opened 2024-11-19
Prevent redundant telemetry on authentication session changes

pastuxso:
WIP...


sourishkrout:
🤔 I believe the idea was to centralize the logic to "flip" LCID into a single command, based on this refactor it appears it's being undone. The "code reuse" argument here is overly correcting on the the code vs command side. I'm not sure that's warranted unless we know it solves the original problem? Does it?

pastuxso:
@sourishkrout, I'll update the description and title according to the latest changes on this PR to answer correctly.

PR #1806, opened 2024-11-19
Open Cloud Panel only when the Save button is clicked

pastuxso:
The Cloud Panel was being opened automatically for auto-saving, which is unnecessary. This change ensures it only opens when explicitly required.


sourishkrout (approved):
:+1: LGTM

Issue #1807, opened 2024-11-19
Issues with `EC2 render->Connect via SSH`


sourishkrout:
1. When Red Hat linux is selected as target, the username detection fails. Should be `ec2-user`.
2. Unless the local config's default region matches the EC2 VMs connecting will fail. We should always pass the `--region` flag as part of generating/inserting the cell.

PR #1808, opened 2024-11-19
Improve OS username resolution for EC2 instances

pastuxso:
- Expanded supported patterns to include rhel and suse-sles images.
- Introduced a defaultUserName (ec2-user) for unsupported or unmatched image patterns.
- Updated resolveOsUserName function to handle null or undefined values for imageName.
- Fixed case-sensitivity issues by converting imageName to lowercase before matching.
- Simplified logic to ensure robust and predictable username resolution.
- Included the **--region** flag

Closes: #1807


sourishkrout (approved):
:+1: LGTM

PR #1822, opened 2024-11-25
Refactor StatefulAuthProvider for static initialization and improved session management

pastuxso:
This refactor centralizes session management, reduces duplication, and improves maintainability across the extension.

### DIFF
--- a/src/extension/api/client.ts
+++ b/src/extension/api/client.ts
@@ -5,21 +5,26 @@ import { Uri } from 'vscode'
 
 import { getRunmeAppUrl } from '../../utils/configuration'
 import { getFeaturesContext } from '../features'
+import { StatefulAuthProvider } from '../provider/statefulAuth'
 
-export function InitializeClient({
-  uri,
-  runmeToken,
-}: {
-  uri?: string | undefined
-  runmeToken: string
-}) {
+export async function InitializeCloudClient(uri?: string) {
+  const session = await StatefulAuthProvider.getSession()
+
+  if (!session) {
+    throw new Error('You must authenticate with your Stateful account')
+  }
+
+  return InitializeClient({ uri, token: session.accessToken })
+}
+
+export function InitializeClient({ uri, token }: { uri?: string | undefined; token: string }) {
   const authLink = setContext((_, { headers }) => {
     const context = getFeaturesContext()
     return {
       headers: {
         ...headers,
         'Auth-Provider': 'platform',
-        authorization: runmeToken ? `Bearer ${runmeToken}` : '',
+        authorization: token ? `Bearer ${token}` : '',
         'X-Extension-Id': context?.extensionId,
         'X-Extension-Os': context?.os,
         'X-Extension-Version': context?.extensionVersion,
--- a/src/extension/commands/index.ts
+++ b/src/extension/commands/index.ts
@@ -32,7 +32,6 @@ import { Kernel } from '../kernel'
 import {
   getAnnotations,
   getNotebookCategories,
-  getPlatformAuthSession,
   getTerminalByCell,
   openFileAsRunmeNotebook,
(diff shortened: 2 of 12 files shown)


pastuxso:
Hi @sourishkrout,

I had to rollback the reactive approach in **cloud.ts** ([d2040b1](https://github.com/stateful/vscode-runme/pull/1822/commits/d2040b1d33064ba30eea09e1bab7a25bff7737b0)). Unfortunately, the subscription isn't being triggered, so the panel fails to refresh its content as intended.

I've recorded a video demonstrating the steps to test the desired functionality. Your expertise would be greatly appreciated to resolve this issue. Let me know if you need any further details! 

Thanks! 🙌

https://github.com/user-attachments/assets/09ae8b93-fe77-4f10-8cb9-7196c504682e

sourishkrout:
@pastuxso When I open the Runme side panel, I'll immediately get prompted to log in. That's not right. It should display the "Sign in or create account" button first.

pastuxso:
Just to confirm, is it happening for Stateful or Runme ext?

sourishkrout:
> Just to confirm, is it happening for Stateful or Runme ext?

I've only tested Runme so far.

src/extension/utils.ts
```diff
@@ -559,16 +559,16 @@ export function convertEnvList(envs: string[]): Record<string, string | undefine
   )
 }
 
-export function getGithubAuthSession(createIfNone: boolean = true) {
-  return authentication.getSession(AuthenticationProviders.GitHub, ['user:email'], {
-    createIfNone,
-  })
-}
-
 export async function getPlatformAuthSession(createIfNone: boolean = true, silent?: boolean) {
```
sourishkrout, line 562: What's this for? Shouldn't all this be absorbed into the singleton? Seems leaky to have things call in and out of "utils"

pastuxso: Good catch. Having seen this, I think we can move it into the Singleton too.

pastuxso: I believe it should be named to something like this: **getOrCreateSession()**

sourishkrout: Sounds good

pastuxso:
> @pastuxso When I open the Runme side panel, I'll immediately get prompted to log in. That's not right. It should display the "Sign in or create account" button first.

I’ve tried to replicate it for Runme, but I couldn’t. The only way to replicate is within Stateful Extension. I wonder if a specific setting might be causing this issue. Would you mind sending me a screenshot of your panel so I can take a closer look?

sourishkrout:
> > @pastuxso When I open the Runme side panel, I'll immediately get prompted to log in. That's not right. It should display the "Sign in or create account" button first.
> 
> @sourishkrout, I’ve tried to replicate it for Runme, but I couldn’t. The only way to replicate is within Stateful Extension. I wonder if a specific setting might be causing this issue. Would you mind sending me a screenshot of your panel so I can take a closer look?

I believe this below is the issue. It forces the creation of a session inside the call stack without 'true' here. That being said, I wonder if the client flag should be an object and explicit. Optional arguments are easy to overlook.

```
diff --git a/src/extension/panels/cloud.ts b/src/extension/panels/cloud.ts
index 4c917942..6eec159b 100644
--- a/src/extension/panels/cloud.ts
+++ b/src/extension/panels/cloud.ts
@@ -59,7 +59,7 @@ export default class CloudPanel extends TanglePanel {
 
     log.trace(`${this.identifier} webview resolving`)
 
-    const html = await this.getHydratedHtml()
+    const html = await this.getHydratedHtml(true)
     webviewView.webview.html = html
     webviewView.webview.options = {
       ...webviewOptions,
```

sourishkrout (approved):
✅ LGTM

PR #1823, opened 2024-11-27
Update VS Code dependency to the latest compatible version

pastuxso:
This PR updates VS Code dependency to the latest compatible version following its FAQ https://code.visualstudio.com/docs/supporting/FAQ#_can-i-run-vs-code-on-older-windows-versions

Depends on https://github.com/stateful/vscode-telemetry/pull/392


sourishkrout (approved):
✅ using new upstream `vscode-telemetry` which includes the same changes.

PR #1825, opened 2024-11-27
Dependabot Updates 20241127

pastuxso:
Ignored **@connectrpc/connect-node** from 1.6.1 to 2.0.0 https://github.com/stateful/vscode-runme/pull/1816, which is not passing local and CI checks, and contains a major upgrade with breaking changes for us.


sourishkrout (approved):
✅ LGTM

PR #1827, opened 2024-11-28
Prevent ghost cells from changing cell focus

jlewi:
* The bug is described in https://github.com/jlewi/foyle/issues/316
* The problem was we weren't deleting the previous ghost cells until we were ready to insert the new ghost cells. When we deleted the ghost cells this could shift the current cells up. This could lead to the active cell shifting up and losing focus. This was very confusing. It felt like the ground was changing underneath you.

* This PR fixes that by deleting the ghost cells from the current compeltion as soon as you change the cell focus to a new cell. Therefore by the time you start typing the previous ghost cells have already been removed.

* I think this is a much better experience.

* I think the original UX was motivated in part to give you time to look at the previous ghost cells while you are still typing. I also think it might have partially been for demo purposes to give you some ghost cells to continue to look at.

* In practice, I find that if I don't select a ghost cell immediately after switching cell focus I never bother to look at it. So persisting the ghost cells after switching focus wasn't adding any value.

* On the other hand, the cell losing focus was super frustrating.

* Fix https://github.com/jlewi/foyle/issues/316

### DIFF
--- a/src/extension/ai/ghost.ts
+++ b/src/extension/ai/ghost.ts
@@ -255,44 +255,67 @@ export class GhostCellGenerator implements stream.CompletionHandlers {
       return
     }
 
-    const cell = getCellFromCellDocument(editor.document)
-    if (cell === undefined) {
-      return
-    }
+    // So if we want to delete any ghost cells when the active editor changes
+    // Then we should loop over all the cells and if its a ghost cell do one of two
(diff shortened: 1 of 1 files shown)


jlewi:
@sourishkrout This is ready for review.

sourishkrout (approved):
✅ LGTM.

PR #1844, opened 2024-12-10
Cut off outputs

peraltafederico:
Serialize the full cell output content when saving a notebook. This is needed because XTerm serialization only returns the content that's visible in the terminal, causing the cell outputs to be saved with "incomplete" content.


sourishkrout (approved):
✅ LGTM. Works: https://cloud.stateful.com/commands/d26edcdb-4490-42d7-b166-e64041726ccb?workspace=8dfb5c96-2c05-45c6-bd02-850b3420d487&workspace=d26edcdb-4490-42d7-b166-e64041726ccb

PR #1862, opened 2024-12-17
Fix escape sequences in prompts

sourishkrout:
Fixes https://github.com/stateful/runme/issues/709.

Prompt messages are unquoted which requires them to be valid shell. Once prompt resolution runs any "values" quoted or not are primarily being used for human-consumption. This is why we're removing backslashes from any form of parentheses.

Doing this on the client-side to keep the Program Resolver pure-shell.


sourishkrout:
Actually never mind. Don't want clients to diverge. Will handle this kernel-side.

PR #1868, opened 2024-12-18
Refactor Output Actions: Auto-Save, Save, Preview & Gist

pastuxso:
This PR introduces significant improvements to the output action buttons:

- Renamed **Session Outputs** to **Preview Outputs**: This button now always previews outputs, regardless of **Auto-Save** status.

- Renamed **Preview Gist** to **Preview & Gist**: This button now previews and gists cell-level outputs.

- Enhanced **Auto-Save**: Now automatically saves cell outputs for logged-in users and continuously writes session output files for logged-out users when enabled.

- Detached **Auto-Save** and **Session Outputs**: Local session outputs no longer require **Auto-Save** to be enabled.


sourishkrout:
When I'm logged out, it appears the session file is now always being written. Can you please confirming you're seeing the same issue on your end @pastuxso?

src/client/components/terminal/saveButton.ts
```diff
@@ -0,0 +1,35 @@
...
+  private handleClick(e: Event) {
+    if (e.defaultPrevented) {
+      e.preventDefault()
+    }
+
+    this.dispatchEvent(new CustomEvent('onClick'))
+  }
+
+  render() {
+    let text = this.signedIn ? 'Save' : 'Save to cloud'
```
sourishkrout, line 23: Uppercase `C` in `Save to Cloud`

sourishkrout:
It also appears that `Preview Outputs` with `Auto-Save` turned off continuously track the execution state in the notebook. This is hard to test because the file's always written as per my initial comment.

sourishkrout (changes requested):
Please see comments.

sourishkrout (approved):
✅ LGTM

PR #1873, opened 2024-12-19
Fix features flags loading

pastuxso:
This update addresses a race condition where feature flags were not reliably available during the extension’s initialization process.


sourishkrout (approved):
✅ LGTM

PR #1875, opened 2024-12-20
Serializer test fix

hotpocket:
The `cell outputs marshaling` test checks an `item.data.type` property which is not valid per the proto.  This test works because the Notebook.clone() method in `serializer.ts/marshalNotebook()` stores this data which is later accessed in the test. During normal operation the NotebookData that marshalNotebook receives has a `Buffer` object for `item.data` not a `{'type':'Buffer','data':[1,2,3,...]}` type.  A fix is provided here to make the fixture more representative of what would be encountered in normal operation of the extension.

Side note:  work I am doing to convert from timostamm-ts to protobuf_es creates a Notebook differently, which strips out this invalid structure/data, which is how this bug was caught. I believe this fix will allow the two to coexist. The other PR for the conversion is almost ready...


sourishkrout (approved):
✅ LGTM. Will merge pending all tests are passing the pipeline.

sourishkrout:
Ran the tests out-of-band and they all passed. The e2e tests suite requires `stateful` org secrets which is why didn't pass for this PR.

PR #1876, opened 2024-12-20
Convert GrpcSerializer to use buildbuf_es instead of timostamm-protobuf-ts

hotpocket:
As I understand it buildbuf_es has the ability to use transports other than grpc.  Looking forward to allowing this to run in the web we need a transport that can be used there, namely a connect transport.  Every attempt has been made to preserve backwards compatibility but there are some unavoidable differences:
* The transport needs to be constructed and needs a cert & key to communicate with a tls server.  To this end the kernelServer.ts was modified as well to expose this data with the goal of avoiding code duplication.
* A force cast was needed in saveCellExecution.ts, presumably due to graphql.ts, with the suspected data types it contains coming from the timostamm proto bindings. This file exists in a `__generated-platform__` directory but I was unable to find how this file was generated and investigate further. 

Let me know what you think.

Thanks!

* https://github.com/stateful/runme/issues/641 documents the motivation to migrate to bufbuild_es


jlewi:
> The transport needs to be constructed and needs a cert & key to communicate with a tls server. To this end the kernelServer.ts was modified as well to expose this data with the goal of avoiding code duplication.

I think this is temporary because right now the Executor is still using timostamm-ts; so we need to construct two different transports 1) one using timostamm-ts and 2) one using es libraries.

This should go away as soon as we migrate the Executor to use the es libraries and then we can go back to the pattern of having the Server just return a transport.

hotpocket:
it seems these tests are failing for the same reason that [my prior PR](https://github.com/stateful/vscode-runme/pull/1875) failed for.  Is this your assessment as well?  I ran the tests locally and got all green lights.

pastuxso:
Hi @hotpocket, the issue is caused by the required permissions needed to run the E2E tests for the GitHub integration.

I’ve opened this [PR](https://github.com/stateful/vscode-runme/pull/1887) to skip those tests specifically for forked repositories.

Let's see what @sourishkrout thinks about this solution.

sourishkrout:
> it seems these tests are failing for the same reason that [my prior PR](https://github.com/stateful/vscode-runme/pull/1875) failed for. Is this your assessment as well? I ran the tests locally and got all green lights.

This is a false alarm. This one 2e2 test fails because branches outside the Stateful org won't have access to a secret required to make this tests pass. Safe to ignore.

sourishkrout:
> Let's see what @sourishkrout thinks about this solution.

Merged into `main` which will definitely help avoid false positives. Thank you!

src/extension/serializer.ts
```diff
@@ -474,36 +478,35 @@ export class WasmSerializer extends SerializerBase {
 }
 
 export class GrpcSerializer extends SerializerBase {
```
sourishkrout, line 480: @hotpocket The existing abstraction that elect to run the serializer via GRPC or WASM (legacy; should be removed). Ideally we'd have a `BufEsSerializer` or something like that which makes it possible to run the serializer using a feature flag.

I'm somewhat uncomfortable making a hard cutover because it requires front-loading of extensive testing which is very difficult to coordinate.

sourishkrout (changes requested):
See my question, please.

src/extension/serializer.ts
```diff
@@ -512,8 +515,45 @@ export class GrpcSerializer extends SerializerBase {
     )
   }
 
-  private async initParserClient(transport?: GrpcTransport) {
-    this.client = initParserClient(transport ?? (await this.server.transport()))
+  private async initClient(): Promise<void> {
```
sourishkrout, line 518: I'd likely be cleaner/easier to have the `KernelServer` supply transports for both GRPC and ES.

src/extension/serializer.ts
hotpocket: This was done successfully [here](https://github.com/hotpocket/vscode-runme/pull/4/commits/1000516849483057346fdbc8cd58b8cba8a5d77a#diff-f28bac02639ff6caf3f3123caf29961e753eb6b74637f5d893a51d935d49f162) as `ConnectSerializer` but was reverted due to the amount of duplicated code.  The serialization logic is very similar, but the structure of the types as they manifest with each proto binding required changes inside many serializer functions (off the top of my head: the marshal methods, save, revive, cache, ...). I tried to construct an intermediary base class (`GrpcSerializerBase`), but there was still significant code that lived in the concrete implementations.  The initial decision was to duplicate code, hence this implementation, but it was deemed to be too much duplication, and a potential maintenance issue. 

The timostamm bindings seem to be assumed throughout the code base, so using ES has been challenging here, and I expect will be elsewhere as we walk the dependencies. I have not been able to find a one size fits all approach so any suggestions are greatly welcomed.

src/extension/serializer.ts
hotpocket: this appears to be a [timostamm](https://github.com/timostamm/protobuf-ts/blob/main/MANUAL.md#grpc-transport) transport used by a timostamm [HealthClient](https://github.com/hotpocket/vscode-runme/blob/4f313c6e21de3731ec4ea7a7a047184aa4bd4009/src/extension/server/kernelServer.ts#L122).  Should I seek out an analogue in the ES space?  How would you suggest making this modification to avoid the smallest cascade of changes due to the dependency graph?

src/extension/serializer.ts
jlewi: Some additional context.

I think in order to eventually support the Connect Protocol we need to support both the connect protocol and the gRPC protocol for some period of time. There are two reasons for this

1. The Server currently only supports the gRPC protocol; so the connect protocol won't be functional until the server supports the connect protocol
2. Doing a hard cutover to connect protocol in both the server and extension is going to be more difficult to coordinate
3. We probably want to bake the connect protocol before dropping support for the gRPC protocol.

The original thinking was to support gRPC and Connect simultaneously we need to have two different Serializer classes using the timostamm and ES libraries respectively.

However, we realized that since the connect libraries support both the gRPC protocol and Connect Protocol we can support both protocols with a single class but letting that class use either a gRPC transport or a connect transport. 

@sourishkrout  regarding

> I'm somewhat uncomfortable making a hard cutover because it requires front-loading of extensive testing 

Are there particular risks you are worried about? 

Since this PR is changing
1. The Generated code used to represent the protos
2. The library used to serialize the protos and gRPC

I can think of two main failure modes
1. A bug in the conversion logic between VSCode data structures and the new proto classes
2. A bug in the connect implementation of the proto & gRPC spec

For #1 the code diff is quite small. So between the small code diff and unittests, how big do you think the risk?

Are you worried about #2? I would have expected the connect libraries to be well tested and vetted since its kind of the whole point of the underlying company.

Is there some other risk you forsee? Does the executor have some dependency on the serializer that isn't being surfaced in the diff?

That said if you want to preserve the existing code path and let the new one bake for it. I think that's pretty straightforward. It should be fairly simple just to restore the existing GRPCSerializer class and introduce an option to select the serializer.

As @hotpocket mentioned the complexity with two classes arose because of an effort to avoid code duplication. I don't think there's any reason to avoid code duplication if we want to preserve the GRPCSerializer. The impetus for minimizing code duplication was that both classes would need to exist for a long time because we won't be able to remove GRPC support anytime soon. However, we only need to keep the GRPCSerializer around long enough to validate that the Connect class is working as intended with gRPC. If that's a reasonable short time (1 week? 2 weeks?) then the burden of supporting duplicate code paths is pretty low.

So @sourishkrout if we preserve GRPCSerializer, how much bake time would you want before you are comfortable removing it. If you and me use the ConnectSerializer for 1 week is that sufficient?

src/extension/serializer.ts
jlewi: @hotpocket I think what @sourishkrout was suggesting is that right now the Kernel has a method [Kernel.transport](https://github.com/stateful/vscode-runme/blob/99911932ee82763f0d0710bf25fcbae8e0a8d030/src/extension/server/kernelServer.ts#L200) which returns a GRPCTransport for the timostamm library. I think he's suggesting using the same pattern to create the transport for the ES library rather than creating the transport inside your serializer.ts

e.g. add a method *kenel.connectTransport()* which returns the transport. Then all the code for initializing the transport can move out of the serializer into the kernel.

jlewi:
@hotpocket @sourishkrout WDYT?

src/extension/serializer.ts
hotpocket: So the suggestion is to add a function to kernelServer to construct and return an es transport (likely identical to what was added in serializer in this PR) but not add any of the monitoring or integration that it does for the timostamm transport. If we cut over to es in the future then the monitoring and integration can be address then IIUC.

src/extension/serializer.ts
hotpocket: @sourishkrout @jlewi I think this approach makes good sense. The changes are small enough to isolate issues should they arise and it avoids maintaining duplicate code paths.

sourishkrout: > So @sourishkrout if we preserve GRPCSerializer, how much bake time would you want before you are comfortable removing it. If you and me use the ConnectSerializer for 1 week is that sufficient?

Looking at it again now. I'll get back shortly.

sourishkrout:
@hotpocket @jlewi did some simple refactoring to get closer to merging. Can we please cherry-pick [this commit](https://github.com/stateful/vscode-runme/pull/1897/commits/49f54b948a42428a1c7bf875ab5deb8cc28e6425) into this branch? I can't use @hotpocket's branch as a base without doing some forking limbo.

src/extension/serializer.ts
sourishkrout: Yes, @hotpocket, essentially, the idea is to modify the `IServer` to have both `this.server.tcpTransport()` and `this.server.connectTransport()`. We don't need the health check to use Connect.

hotpocket:
I have merged your cherry-picked code.  I also moved the transport method to kernelServer and followed the patterns found there for the pre-existing transport, minus the health status checks. 

Some new code has caused a coverage failure at 78.93% of 79%.  If you'd like me to write some tests let me know.  I looked at the kernelServer.test.ts to see if I should add tests for the new connect transport method but didn't see any prior transport tests to model it on.  Let me know how you want to proceed WRT testing coverage and anything else you find.

src/extension/serializer.ts
jlewi: LGTM

jlewi:
So it looks like we agree on the plan. It looks like the path forward is as follows

1. Merge PR(https://github.com/stateful/vscode-runme/pull/1897) which refactors the imports
2. Then Rebase/Merge main to cleanup the diff
3. Restore the GRPCSerializer on this branch
4. Expose a VSCode option to let you select the serializer you want to use

src/extension/server/kernelServer.ts
```diff
@@ -210,6 +222,54 @@ class KernelServer implements IServer {
...
+
+    if (protocol === 'grpc') {
+      this.#connectTransport = createGrpcTransport(<GrpcTransportOptions>{
+        ...transportOptions,
+        httpVersion: '2',
+      })
+    } else {
+      this.#connectTransport = createConnectTransport(<ConnectTransportOptions>{
+        ...transportOptions,
+        httpVersion: '1.1',
```
jlewi, line 266: Should this be 2?

hotpocket: it depends on what the server it's communicating with wants. the [fake_server](https://github.com/hotpocket/vscode-runme/blob/v1/fake_server/con_serv.ts) I was using as my first stab at this process was using 1.1, but if you construct a fastify instance like this `fastify({http2:true})` you can switch it to 2 in the [client](https://github.com/hotpocket/vscode-runme/blob/v1/fake_server/client.ts) and it won't error out.  Since we don't have a server yet that talks connect, I can switch it to 2 in anticipation if that's your vote. This code here is a leftover of the last "good" value.

hotpocket: since it will make the code shorer, and we don't have a server that talks connect at the moment I'll eliminate this conditional and assume we'll have http2 in the end.

src/extension/server/kernelServer.ts
```diff
@@ -395,6 +454,7 @@ class KernelServer implements IServer {
     await this.acceptsConnection()
 
     this.#onTransportReady.fire({ transport: await this.transport(), address: this.address() })
+    this.#onConnectTransportReady.fire(await this.connectTransport())
```
sourishkrout, line 457: I think it'd be prudent to match connect's type `{ transport, address }` as per the line above

src/extension/server/kernelServer.ts
```diff
@@ -210,6 +222,53 @@ class KernelServer implements IServer {
     return this.#transport
   }
 
+  async connectTransport(protocol: 'grpc' | 'connect' = 'grpc'): Promise<ConnectTransport> {
+    if (this.#connectTransport) {
+      return this.#connectTransport
```
sourishkrout, line 227: We can't use a cache/LRU strategy because the transport will no longer work if the underlying kernel server restarts. Easy test is to send a `kill -9 <pid>` to the kernel server running for the respective window (as per `Output->Runme` log line toward the top).

src/extension/serializer.ts
```diff
@@ -474,35 +481,39 @@ export class WasmSerializer extends SerializerBase {
...
+    protected server: KernelServer,
     kernel: Kernel,
   ) {
     super(context, kernel)
 
     this.togglePreviewButton(GrpcSerializer.sessionOutputsEnabled())
 
+    // cleanup listener when it's outlived its purpose
     this.ready = new Promise((resolve) => {
       const disposable = server.onTransportReady(() => {
```
sourishkrout, line 507: I believe `server.onTransportReady(...)` should be `server.onConnectTransportReady(...)` since the `GrpcSerializer` will no longer use TCP Transport. This likely only "works" because both transports are firing simultaneously. We should use Connect to avoid foot-guns when we sunset the TCP Transport.

sourishkrout (changes requested):
Please see review, @hotpocket @jlewi.

Serialization is stateless on the API-level. We should make sure it works uninterrupted when the kernel server restarts. This is how the TCP transport works.

It's "easy" to test by sending `kill -9 <pid>` to the kernel's process. The `pid` is in `Output->Runme`.

src/extension/server/kernelServer.ts
hotpocket: @sourishkrout AFAICT I can't use the `address()` value to make a connect connection because it might be a unix socket like "unix://socket", at least I have avoided using it in `connectTransport` for this reason.  It is for this reason that I modified the type of the event, to make this restriction explicit. I'm happy to change it for consistency reasons, but wanted to point out my rationale for this intentional change to double check my thinking because of how i obtain an address in `connectTransport`

src/extension/serializer.ts
hotpocket: fixed. ty

src/extension/server/kernelServer.ts
sourishkrout: Yes, @hotpocket. I meant the comment in a general sense. Any code path that assumes a static/cached transport will interfere with gracefully handling kernel restarts.

src/extension/server/kernelServer.ts
sourishkrout: That's fine, @hotpocket. Outside the Runner (out of scope), the `address` is mainly used for reporting purposes, such as logging. Since the ready-made transport is part of the event payload, there's no need to construct one off the address in the consumer/receiver.

src/extension/server/kernelServer.ts
sourishkrout: I believe HTTP2 is required for streaming RPCs. Unary RPCs likely get away with HTTP1.1.

However, let's require HTTP2 since the Kernel server supports it anyways.

sourishkrout:
@hotpocket here's another path to apply to prevent breaking the Stateful product integration. Sorry for the poor man's PR but I'm trying to get around fork cloning dance.

```
diff --git a/src/extension/messages/platformRequest/saveCellExecution.ts b/src/extension/messages/platformRequest/saveCellExecution.ts
index 1e370f89..05f4c1ff 100644
--- a/src/extension/messages/platformRequest/saveCellExecution.ts
+++ b/src/extension/messages/platformRequest/saveCellExecution.ts
@@ -4,7 +4,7 @@ import { Uri, env, workspace, commands } from 'vscode'
 import { TelemetryReporter } from 'vscode-telemetry'
 import getMAC from 'getmac'
 import YAML from 'yaml'
-import { FetchResult, MutationOptions } from '@apollo/client'
+import { FetchResult } from '@apollo/client'
 
 import { ClientMessages, NOTEBOOK_AUTOSAVE_ON, RUNME_FRONTMATTER_PARSED } from '../../../constants'
 import { ClientMessage, FeatureName, IApiMessage } from '../../../types'
@@ -18,9 +18,11 @@ import { InitializeCloudClient } from '../../api/client'
 import {
   CreateCellExecutionDocument,
   CreateCellExecutionMutation,
+  CreateCellExecutionMutationVariables,
   CreateExtensionCellOutputDocument,
   CreateExtensionCellOutputMutation,
   CreateNotebookInput,
+  MutationCreateExtensionCellOutputArgs,
   ReporterFrontmatterInput,
 } from '../../__generated-platform__/graphql'
 import { Frontmatter } from '../../grpc/parser/tcp/types'
@@ -136,7 +138,7 @@ export default async function saveCellExecution(
       // TODO: Implement the reporter to normalize the data into a valid Platform api payload
       const mutation = {
         mutation: CreateExtensionCellOutputDocument,
-        variables: {
+        variables: <MutationCreateExtensionCellOutputArgs>{
           input: {
             extension: {
               autoSave: autoSaveIsOn,
@@ -188,7 +190,8 @@ export default async function saveCellExecution(
           },
         },
       }
-      const result = await graphClient.mutate(mutation as MutationOptions)
+
+      const result = await graphClient.mutate(mutation)
       data = result
     }
     // TODO: Remove the legacy createCellExecution mutation once the reporter is fully tested.
@@ -238,9 +241,9 @@ export default async function saveCellExecution(
       }
       const sessionId = kernel.getRunnerEnvironment()?.getSessionId()
 
-      const result = await graphClient.mutate({
+      const mutation = {
         mutation: CreateCellExecutionDocument,
-        variables: {
+        variables: <CreateCellExecutionMutationVariables>{
           input: {
             stdout: terminalContents,
             stderr: Array.from([]), // stderr will become applicable for non-terminal
@@ -287,7 +290,9 @@ export default async function saveCellExecution(
             },
           },
         },
-      })
+      }
+
+      const result = await graphClient.mutate(mutation)
 
       data = result
     }
```

sourishkrout:
> Some new code has caused a coverage failure at 78.93% of 79%. If you'd like me to write some tests let me know. I looked at the kernelServer.test.ts to see if I should add tests for the new connect transport method but didn't see any prior transport tests to model it on. Let me know how you want to proceed WRT testing coverage and anything else you find.

I can handle the test coverage stuff once we're closer to merging. Give that little drop; not something to worry about right now, @hotpocket.

src/extension/server/kernelServer.ts
hotpocket: fixed.

src/extension/server/kernelServer.ts
```diff
@@ -210,6 +220,45 @@ class KernelServer implements IServer {
...
+        if (!getTLSEnabled()) {
+          return {}
+        }
+        const pems = await KernelServer.getTLS(this.getTLSDir())
+
+        return {
+          nodeOptions: {
+            key: pems.privKeyPEM,
+            cert: pems.certPEM,
+            rejectUnauthorized: false,
```
sourishkrout, line 235: @hotpocket, why is `rejectUnauthorized` set to `false`? Is this intentional? Looks alarming without a comment explaining why.

sourishkrout: The Kernel server connection is secured using mutual authentication, so only identified parties accessing the cert+private key can talk to one another. This is a core security feature.

sourishkrout (changes requested):
Let's clarify my comment about `rejectUnauthorized: false`

src/extension/server/kernelServer.ts
sourishkrout: I guess that `pems.certPEM` should be passed as both `cert` and `ca` because it's essentially a self-signed cert the Runme kernel rotates every 30d.

Thanks to the structural overlap passing the compiler, it looks like `addTlsConfigIfEnabled` gets by without strong typing. Security is a top priority, so let's harden this function by explicitly specifying the Option(s) types.

Having part of the object construction in lines 243-251 complicates this. Specifically, `httpVersion: '2'` branches into a union type that is "undecided" to the upstream `addTlsConfigIfEnabled`. If the types were specified, this would be illegal, and the compiler would error.

sourishkrout:
@hotpocket @jlewi Runme's release queue is pretty much "flushed", so early next week, there'd be a time to get this into a pre-release for final testing.

src/extension/server/kernelServer.ts
hotpocket: Thanks @sourishkrout , I had considerable difficulty with the tls part of this and apparently hadn't completely gotten it right.  I think your concerns are addressed with the latest "self-signed cert ..." commit, if not let me know what still needs to be done.

hotpocket:
> Let's clarify my comment about `rejectUnauthorized: false`

`rejectUnauthorized: false` was to get the queries to not be rejected. I changed this to `ca: pems.certPEM` and it runs just the same.  See prior notes & commit.

src/extension/server/kernelServer.ts
sourishkrout: Looks excellent, @hotpocket! You addressed all my points. It's much cleaner and easier to grok now.

I wish TLS, especially outside of Web PKI ("https" with the lock in the browser), was more straightforward.

sourishkrout (approved):
✅ LGTM. I will merge this on Monday and make sure it plays nicely with the Cloud Product before releasing it as a `pre-release` of the Runme VS Code extension.

jlewi:
@sourishkrout @hotpocket Woo Hoo.  I can also install it. I pretty much live in Runme these days.

jlewi (approved):

sourishkrout:
All tests passed locally. I'm doing more final testing on `main` before shipping a pre-release later today or early tmrw.

sourishkrout:
> I pretty much live in Runme these days.

Music to my ears, @jlewi.

sourishkrout:
I just published a `pre-release` with this PR, `3.11.1737499931` (https://github.com/stateful/vscode-runme/releases/tag/3.11.1-edge.0), on the VS Code marketplace, @hotpocket + @jlewi.

Let's please make an effort to use it this week to break it in ahead of a `stable` release. Here's how to switch to it: https://docs.runme.dev/installation/vscode#bleeding-edge-features

jlewi:
Installed it and will use it. 
Screenshot 2025-01-21 at 3 13 20 PM

sourishkrout:
One problem I'm just now realizing is that when the Parser's exclusively running on GRPC-Web, we can no longer use a Unix Domain Socket. The latter is not an uncommon choice if corporate firewalls block sockets on loopback.

Unfortunately, GRPC-TCP and GRPC-Web appear to be mutually exclusive, which probably explains why there's a `protobuf-ts` and a `protobuf-es`.

🤔 Trying to determine the best course of action if anybody has ideas.

sourishkrout:
> One problem I'm just now realizing is that when the Parser's exclusively running on GRPC-Web, we can no longer use a Unix Domain Socket. The latter is not an uncommon choice if corporate firewalls block sockets on loopback.
> 
> Unfortunately, GRPC-TCP and GRPC-Web appear to be mutually exclusive, which probably explains why there's a `protobuf-ts` and a `protobuf-es`.
> 
> 🤔 Trying to determine the best course of action if anybody has ideas.

I decided to roll forward instead of rolling back. Work in progress is here @hotpocket + @jlewi https://github.com/stateful/vscode-runme/pull/1927.

jlewi:
Why do we need to exclusively support grpc-web?

The connect library supports 
- grpc
- grpc-web
- connect protocol

The only thing that changes is the transport.
So my assumption was that we'd continue to support grpc indefinitely.

sourishkrout:
> Why do we need to exclusively support grpc-web?
> 
> The connect library supports
> 
> * grpc
> * grpc-web
> * connect protocol
> 
> The only thing that changes is the transport. So my assumption was that we'd continue to support grpc indefinitely.

Can you point me at docs or an example where `protobuf-es` is being used with a regular TCP transport and client? I have not been able to find anything.

What I'm gathering from, e.g., https://connectrpc.com/docs/node/using-clients#grpc, is that while the plain-old `grpc` protocol runs over a HTTP transport, there are no longer TCP/UDS options.

So, in a nutshell, the issue I'm seeing is that the protocol + transport (as in L7 vs L4) combos exclude UDS. Supported
* grpc over HTTP(S)
* grpc-web over HTTP(S)
* connect protocol over HTTP(S)

To get around the local firewalls, we need something that accepts `unix:///tmp/runme.sock` which `protobuf-ts` supports because it layers on L4 TCP. It seems silly to me that the `Connect` library does not support L4 TCP. Even if it's considered "legacy" it makes migrations super hard.

 Perhaps I misunderstand, @jlewi?

sourishkrout:
This issue seems to confirm my findings: https://github.com/connectrpc/connect-es/issues/756

jlewi:
@sourishkrout Thanks for the explanation. 

> Perhaps I misunderstand, @jlewi?
Nope; it was I who misunderstood. It looks like there is a distinction between "gRPC protocol" and "gRPC features". If I understand this [comment](https://github.com/connectrpc/connect-es/issues/756#issuecomment-1699672077). It looks like support for sockets was a feature of the particular gRPC client library and not specified by the protocol. I'm assuming "every gRPC feature" is referring to every feature of the "official" grpc implementation.

It sounds like you may have found a workaround. If that doesn't work another approach might be to maintain separate implementations but use the Connect generated classes for the generated datastructures.

The reason we got code duplication between a Serializer using protobf-ts and a serializer using protobuf-es was because the generated, typed classes for the protos were different. The wire serialization format should be exactly the same though. So there's probably an implementation where we use the protobuf-es Classes everywhere and then when making the request you use protobuf-ts "grpc client".  Ideally we wouldn't need to do an extra round trip of serialization e.g.
NotebookEs -> bytes -> NotebookTs -> ts client lib.

sourishkrout:
👍 @jlewi. The docs are anything but clear to be fair. The HTTP at L7 versus TCP at L4 could be more prominent.

> It sounds like you may have found a workaround.

Good news! I managed to shove a UDS transport under the connection. Works locally. I have high hopes that it'll work on Linux, too, since these are all standard Node.js APIs.

I should be able to encapsulate the details behind what amounts to a custom transport and combine it with my refactor branch which is much cleaner. If it all works it's a win-win 🤞.

sourishkrout:
> I have high hopes that it'll work on Linux, too, since these are all standard Node.js APIs.

Success! Works on Linux, too 🥳. Integrated in https://github.com/stateful/vscode-runme/pull/1927. Continuing the discussion over there.

PR #1898, opened 2025-01-10
Debounce keystrokes in AI event stream using obeservables

sourishkrout:
The idea is to vastly simplify and allow more control how the UX behavior impacts generation/streaming, e.g. debounce, without an "imperative implementation".

Work in progress:

- [x] Retro-fit and fix existing tests for streaming
- [x] Slot in debouncing of events with observables ahead of streaming
- [x] Ensure contextIDs are valid between text changes and execution

Mostly notably, this will fix where the first key stroke, e.g. `say hello in the terminal`, in the cell will produce a "I can't help with "s"..." kind of response.


sourishkrout:
@jlewi before I go tackle the upstream implementation, could you please review `'manual foyle streaming RPC test'` in `streaming.test.ts` to make sure I fixed it correctly. Thank you!

jlewi:
>  I go tackle the upstream implementation

Does this mean the test is failing and your wondering if its an issue in the test or upstream? What is upstream here? Foyle? Observable?

I will try to run it after you respond and see what's going on it.

> How  are you planning on using Observable here? I found it difficult wrap my mind around how to wrap the iterators that the connect rpc library uses with the observable framework. In particular, how to cancel a stream and start a new one when a user switches cells.  Foyle's protocol is that a new streaming request should be initiated and the current one closed whenever the context (basically the selected cell) changes.

> Mostly notably, this will fix where the first key stroke, e.g. say hello in the terminal, in the cell will produce a "I can't help with "s"..." kind of response.

One tricky bit we should keep in mind is making sure completions are correctly triggered on code cells when they are executed.

src/extension/ai/stream.ts
```diff
@@ -20,7 +32,7 @@ export interface CompletionHandlers {
   // If null is returned then no request is generated
   buildRequest: (
     cellChangeEvent: CellChangeEvent,
-    firstRequest: boolean,
+    options: { firstRequest: boolean; handleNewCells: boolean },
```
jlewi, line 35: Can you explain this change? What is the meaning of handleNewCells?

jlewi:
Thanks for the clarification and for updating the test. Sorry for the slow reply. 

So it looks like the test has fallen behind main in at least the following ways

1. StreamCreator constructor needs a client argument
2. Set a contextId on the requests

Does the test also timeout for you?  It looks like the test depends on the response containing "done" 
https://github.com/stateful/vscode-runme/blob/57b16b19feb3dd6e6db04877f6170ba8b0beb322/tests/extension/ai/streaming.test.ts#L106
to now the end of the stream was reached.

I have some recollection that I was using a fake server that just echoed back whatever the client sent. 
I couldn't find any mention of that though in my notes; or a copy of the server. 

I think it would be easy to code something up to be used for the purposes of testing. I guess it depends on how we intend to use the test going forward?

src/extension/ai/stream.ts
sourishkrout: It's hacky and not final yet, I introduced it to be able to reuse the `buildRequest` method to control my debounce logic. My plan was to rewrite once I have both the debounce and the tests working end-to-end. I'll let you know when it's ready for a final review.

sourishkrout:
> Does the test also timeout for you?
> 
> It looks like the test depends on the response containing "done"
> 
> https://github.com/stateful/vscode-runme/blob/57b16b19feb3dd6e6db04877f6170ba8b0beb322/tests/extension/ai/streaming.test.ts#L106
> 
> 
> to now the end of the stream was reached.
> I have some recollection that I was using a fake server that just echoed back whatever the client sent. I couldn't find any mention of that though in my notes; or a copy of the server.

Yes. I had to stop here on Saturday when I realized that the tests fails based on the the final response containing `done` or not. I have an idea how to fix this. I think we should keep this "manual" test around.

> 
> I think it would be easy to code something up to be used for the purposes of testing. I guess it depends on how we intend to use the test going forward? A simpler solution might be to count the number of times processResponse has been invoked? I think it should be invoked N times where N is the number of input data that are sent. I probably didn't come up with a better solution that wasn't dependent on the backend echoing back the inputs because I was operating at the limits of my typescript competence

That makes sense. I'm planning to extract a version of the fully integrated "manual" test that uses faking/mocking to verify that the logic remains intact.

sourishkrout:
> The implementation is working but the test's failing. I found it difficult to "optimize" the flow of events because pusher/iterator are "coupled". I have for the moment skipped a rewrite of `PromiseFunctions` and `PromiseIterator` because I realized that we will be using 3 different RPC Transports in Runme: TCP, Connect/JSON, and gRPC-Web. I'd prefer if we standardized on gRPC-Web, but this largely concerns the integration with Foyle.

Renamed the PR to be framed around the debouncing of keystrokes. I'm hesitant to extend observables for all of the event streaming until we have clarity on what gRPC transport we want to use longer term.

src/extension/ai/stream.ts
sourishkrout: @jlewi reverse this change as part of a refactor for clarity

sourishkrout:
Ready for review, @jlewi. For the "functional impact" please see the video above.

jlewi (approved):
One thing to keep in mind with debouncing is that in a lot of cases there is more than the current cell in the notebook. So when you press "s" there might be information in the previous cells that allow the model to make a good guess about what to generate.

It probably still makes sense to do some debouncing/infra on the client side to make good use of the connection.

sourishkrout:
> One thing to keep in mind with debouncing is that in a lot of cases there is more than the current cell in the notebook. So when you press "s" there might be information in the previous cells that allow the model to make a good guess about what to generate.

 👍 these scenarios are covered. A single character (`s`) or even incomplete typing (`say hi in th`) will still be logged if a typing delay of >500ms occurs in between keystrokes. The smart debounce logic aims to have zero loss of information fidelity.

PR #1908, opened 2025-01-15
Flip Reporter API feature flag to default enabled

sourishkrout:
We've tested this in pre-release since early 2024Q4.


peraltafederico (approved):
lgtm!

PR #1910, opened 2025-01-15
Added Command: `runme.authenticateWithStateful`

pastuxso:
This update introduces the runme.authenticateWithStateful command, enabling authentication independently of the Cloud Panel.

Previously, users could not authenticate when the Cloud Panel was unavailable. This command resolves that limitation, providing a direct way to authenticate with Stateful.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -304,15 +304,15 @@
       },
       {
         "command": "runme.authenticateWithGitHub",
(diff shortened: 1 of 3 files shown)


sourishkrout (approved):
✅ LGTM

PR #1913, opened 2025-01-16
Adds support to manually refresh launcher

pastuxso:
The new launcher doesn't have a way to synchronize new, updated, or removed cells/notebooks. For now, this PR introduces a button to manually update the content of the panel. However, in a next iteration, a file system watcher would be considered.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -265,6 +265,12 @@
         "icon": "$(debug-collapse-all)"
       },
       {
+        "command": "runme.refreshTasks",
+        "category": "Runme",
+        "title": "Refresh tasks",
+        "icon": "$(sync)"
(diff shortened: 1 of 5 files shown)


sourishkrout:
While this works functionally and does what we want, I will rewrite it to use Observables. Mediating between Promises vs. Observables is a bit awkward.

pastuxso:
> While this works functionally and does what we want, I will rewrite it to use Observables. Mediating between Promises vs. Observables is a bit awkward.

That works for me. You can make changes whenever you'd like.

sourishkrout (approved):
✅ LGTM

PR #1914, opened 2025-01-17
Review feedback

sourishkrout:
Just some small things:

1. Use rx observables throughout.
2. Sort lexically on the same dir level.


pastuxso (approved):
LGTM 🚀

PR #1915, opened 2025-01-21
Use Buffer instead of Uint8Array

sourishkrout:
This change is required since Web gRPC, unlike TCP gRPC, returns `Buffer` and not `Uint8Array` as Buffer data type.

Required to release https://github.com/stateful/vscode-runme/pull/1876.


pastuxso (approved):
LGTM 🚀

PR #1919, opened 2025-01-22
Fix Preview Outputs when Life Cycle Id is disabled

pastuxso:
The NotebookData structure doesn't include transient metadata, but I need the cell ID to properly track terminal outputs.

I extract the cell ID from the terminal output metadata, which is always available when the cell has a terminal output. This is a more reliable way to get the cell ID

This is particularly problematic when lifecycleIdentity is NONE since cells won't have IDs assigned directly.

Closes: stateful/runme#728

- [x] Ensure session outputs when Life Cycle Id is NONE
- [x] Unmask / Mask is still linked with the Auto-Save
- [x] Preview & Gist is still linked with the Auto-Save


sourishkrout (approved):
✅ LGTM.

PR #1929, opened 2025-01-25
Use RingBuffer to prevent ever-growing buffer

sourishkrout:
Ever-growing buffers are dangerous for memory consumption and downstream API limits. To avoid running into API payload limits, use `scrollback` settings as a buffer limit.

I have encountered issues when a chatty background process runs inside a notebook terminal. After a while, it will experience both remote and kernel API size limitations.

Using the `scrollback` setting will put control into a user's hands.

### DIFF
--- a/src/extension/terminal/terminalState.ts
+++ b/src/extension/terminal/terminalState.ts
@@ -2,6 +2,7 @@ import { Terminal as XTerm } from '@xterm/headless'
 
 import { OutputType } from '../../constants'
 import { RunnerExitReason } from '../runner'
+import { RingBuffer } from '../../ringBuffer'
 
 export type NotebookTerminalType = 'xterm' | 'local'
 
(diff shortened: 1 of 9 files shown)


sourishkrout:
Please review @peraltafederico when you have a moment.

src/extension/terminal/terminalState.ts
```diff
@@ -54,14 +59,16 @@ export class XTermState implements ITerminalState {
...
     } else {
-      this.buffer = this.buffer + new TextDecoder().decode(data)
+      this.buffer.push(this.textDecoder.decode(data))
     }
   }
 
-  cleanBuffer(): void {
-    this.buffer = ''
+  resetBuffer(): RingBuffer<string> {
+    // capactiy is items vs lines because we don't handle line breaks
```
peraltafederico, line 69: @sourishkrout Typo of "capacity"

peraltafederico (approved):
@sourishkrout I loved the RingBuffer!

Just one thing to consider: the capacity of the buffer will generally store more data than what’s rendered to the user based on the rows + scrollback configuration. You’ve probably already thought of this, but I figured it was worth mentioning.
 
 LGTM :+1:

PR #1930, opened 2025-01-27
Remove unnecessary decoding of base64 data

pastuxso:
The GraphQL API returns the decoded data, so we don't need to decode it again.

Depends on: stateful/platform#1296


sourishkrout (approved):
✅ LGTM

PR #1931, opened 2025-01-27
Use workspace folders as cwd for remote notebooks

pastuxso:
Workspace folders now serve as the current working directory for remote notebooks. This is necessary for shebang cell runs.

When no workspace is detected, it defaults to using a temporary directory.


sourishkrout (commented):
✅ LGTM

PR #1861, opened 2024-12-16
Refactor StatefulAuthProvider for static initialization and improved session management

pastuxso:
This refactor centralizes session management, reduces duplication, and improves maintainability across the extension.

- Centralized session change notifications using a new SessionChangeHandler implemented with RxJS and a singleton pattern.
- Refactored StatefulAuthProvider to follow a singleton pattern, simplifying its lifecycle and reducing dependencies.
- Added functionality to retrieve sessions from secrets storage, avoiding potential infinite loops when checking sessions inside a session change listener.
- Removed Kernel and RunmeUriHandler dependencies from StatefulAuthProvider for better modularity.
- Introduced a new GraphQL client instantiator to encapsulate implementation details like authentication token configuration.
- Ensured Webview refreshes correctly on login/logout events.

PR #1932, opened 2025-01-28
Fix for StatefulAuthProvider's change in behavior

sourishkrout:
SignedIn no longer reports cell runs. This fixes it.

I believe this broke with https://github.com/stateful/vscode-runme/pull/1861.


pastuxso (approved):
LGTM 🚀

PR #1947, opened 2025-02-05
First class dagger shell

sourishkrout:
This PR integrates with the new Kernel APIs to transform a notebook into a working Dagger Shell script on-demand.

Currently this will only happen if the notebook's shell is set to `dagger shell` as part of the frontmatter. In a separate PR we might also default based on file extensions (e.g. `.dag`).

This PR requires https://github.com/stateful/runme/pull/736 with local-gen protos to work. An example notebook is committed here: https://github.com/stateful/vscode-runme/pull/1947/commits/7ed343ca21b35bf9294036d709734c966423bf6e.

PS: Tests might fail. Will work on fixing them.


pastuxso (approved):
Everything works fine locally to me 🚀

PR #1956, opened 2025-02-11
ANSI sequences need to be interpreted for TUIs

sourishkrout:
We're coming full circle here. While the `RingBuffer` worked well to manage capacity, it lacks the interpretation of TUI-like ANSI sequences. TUIs ncurses-like UIs that aren't just growing text output, they use the screen more like a canvas.

This PR handles TUIs better (still needs to capture notebook terminal dimensions), especially when saved as Gist or into the Cloud.

Compare https://gist.github.com/sourishkrout/dbc9037bce8fcf17ba283db939ddc804 (no term emu) vs https://gist.github.com/sourishkrout/7d382f162313d01cc9b1ae0120d93e49 (with term emu).

### DIFF
--- a/src/extension/executors/runner/index.ts
+++ b/src/extension/executors/runner/index.ts
@@ -40,7 +40,7 @@ import {
   getServerRunnerVersion,
   isNotebookTerminalEnabledForCell,
 } from '../../../utils/configuration'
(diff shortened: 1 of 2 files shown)


peraltafederico (approved):
LGTM

PR #1957, opened 2025-02-11
Buffering for non-tty output (non-interactive) in the vscode-integrated terminal

sourishkrout:
This mimics what we do in the notebook and makes the terminal output more readable. Especially if both `stdout` and `stderr` are interspersed.

The difference is **with buffering**:
image

Before the change **without buffering**:
image

Example cell (`interactive: false`): https://gist.github.com/sourishkrout/d61c3226473afdc030951bdc167d15cd#file-01jkvp47gcjh2s6mgysf16j06b-md

### DIFF
--- a/src/extension/extension.ts
+++ b/src/extension/extension.ts
@@ -500,7 +500,7 @@ export class RunmeExtension {
       kernel
         .runProgram('echo $SHELL')
         .then((output) => {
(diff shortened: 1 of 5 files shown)


pastuxso (approved):
At my end it works!

PR #1963, opened 2025-02-14
Disable deactivate-panels workflow for edge releases

pastuxso:
This pull request includes a minor update to the `.github/workflows/release.yml` file. The update adds a condition to ensure that the deactivate-panels workflow only runs if the releaseChannel is not set to 'edge'.

This change is necessary to enable and test the Cloud Search and Chat features within the Edge Release channel.


sourishkrout (approved):
✅ LGTM

PR #1975, opened 2025-02-22
Use vscode's withProgress API to inform user AI's thinking

sourishkrout:
Just a visual cue, really, but I think it helps elevate confusion with LLM latency.

Default enabled in feature flags. Can be disabled via `"runme.features.AIProgress": false,` in `settings.json`.

https://github.com/user-attachments/assets/7c1e2b08-db27-417d-be65-b523b995f7f3

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,15 @@
   "main": "./out/extension.js",
   "runme": {
     "features": {
+      "AIProgress": {
+        "enabled": true,
+        "conditions": {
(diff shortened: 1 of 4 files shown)


jlewi:
That's very cool.

sourishkrout:
Cool. Enabling this per default. It can be disabled with a feature toggle as per the PR's description if it causes issues, @jlewi.

PR #2022, opened 2025-03-13
Move from stateful to runmedev org as part of CNCF sandbox admission

sourishkrout:
Search/replace old with new sites.


admc (approved):
👍

PR #2195, opened 2025-09-18
Use XDG runtime dir for UDS sockets

twz123:
Storing the socket in `XDG_RUNTIME_DIR` places it in the user-specific, permission-locked runtime directory instead of the shared `/tmp` directory. This is more secure and adheres to the freedesktop specification.


sourishkrout:
One aspect, I will have to check is whether or not the Electron "sandbox" allows writes to these directories. When we first built the UDS vs TCP feature, I do remember we had trouble with writes being restricted. Turned out `/tmp` was exempt from this. Platforms to test would be: Linux, MacOS, WSL2 on Windows.

